PR #2, opened 2022-04-27
Viz and model output

turnerm:
Some vis and model output for a meeting tomorrow.

@Tinkaa I can't for the life of me get a dual y axis to work to include the CERF data on the plots, maybe you have the wizardry that I need.

### DIFF
--- a/analysis/06_visulaization.md
+++ b/analysis/06_visulaization.md
@@ -1,0 +1,130 @@
+---
+jupyter:
+  jupytext:
+    formats: ipynb,md
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.3'
+      jupytext_version: 1.13.7
+  kernelspec:
+    display_name: global-monitoring
+    language: python
+    name: global-monitoring
+---
+
+# Visulaization
+
+Some viz for the first CERF meeting
+
+```python
+%load_ext jupyter_black
+```
+
+```python
+from pathlib import Path
+
+from datetime import date
+import altair as alt
+import pandas as pd
+
+from src import utils
+```
+
+```python
+OUTPUT_DATA_DIR = Path("../data/output")
+```
+
+```python
+# CERF data
+df_cerf = utils.get_cerf_hdx_data()
+df_cerf = df_cerf[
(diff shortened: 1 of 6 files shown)


analysis/06_visulaization.md
```diff
@@ -0,0 +1,122 @@
...
+```python
+# Get model data, conflict for now
+df_idmc = utils.get_idmc_data()
+df_idmc = df_idmc.loc[df_idmc["displacement_type"] == "Conflict"]
+df_idmc_model = utils.get_model_input_data(df_idmc)
+df_final = utils.run_model(df_idmc, df_idmc_model)
+```
+
+```python
+def plot_country(country, iso3):
```
Tinkaa, line 55: Ahaha I spent waaay to long trying to figure out why the dual axis was not working. In the end it is cause you have to define `properties` on the combined chart #arghh 
Here an example which works 

```
def plot_country(country, iso3):
    base_idmc = (
        alt.Chart(df_final)
        .mark_circle()
        .encode(
            x=alt.X("displacement_date", title="Displacement date"),
            y=alt.X(
                "figure",
                title="Displaced persons",
            ),
            color=alt.Color("Anomaly:N", legend=None),
            tooltip=["standard_popup_text"],
        )
    )

    base_cerf = (
        alt.Chart(df_cerf)
        .mark_point(shape="wedge", color="green", angle=180)
        .encode(x="dateUSGSignature", y="totalAmountApproved")
    )

    chart = (
        (
            base_idmc.transform_filter(alt.datum.iso3 == iso3)
            + base_cerf.transform_filter(alt.datum.countryCode == iso3)
        )
        .properties(title=country)
        .resolve_scale(y="independent")
        .properties(height=100, width=400)
        .configure_axisX(labelAngle=90)
    )

    #     chart.save(OUTPUT_DATA_DIR / f"{country}.png", scale_factor=2)
    return chart
```

Tinkaa: On a side note, I do think the chart is still not suuper clear. I would maybe just plot the displacement data as line chart (or is there too many dates without data, if so could you assume them to be 0?) and then indicate with dots where CERF fundings were (I think the amount is not even that important).

PR #15, opened 2023-07-20
Switch backend to use the Sheets API

caldwellst:
Alright, so I have adjusted the backend to use the Sheets API.  I am going to just cancel the current run in main for today so it doesn't run on the CSVs today as well, I have run it this morning on the Google Sheets and all seems to be moving successfully.

https://github.com/OCHA-DAP/pa-global-monitoring/actions/runs/5608247298

Once we merge, I will remove all of the old CSV files from the drive.


zackarno:
Code looks good to me. Just two quick questions:

- As we do not really need to rely on the `{googlesheets}` for the auto-emails I was wondering if there might be any point in preserving the `{googledrive}` workflow in addition to the `{googlesheets}` (for the dashboard)?
- Have you checked that the main URLS BI will access are not changing on update or do we need to wait and see?

If the links are staying consistent it looks all good to merge to main!

caldwellst:
Links are looking consistent now! Had another struggle with the GitHub Actions build, ended up needing to update `apt-get` prior to installing the build dependencies, and now it looks like everything is working well.

I think for now best to just leave it all as Google Sheets. The only consistently growing sheets that are causing significantly slower performance are the sheets that are having to be shared with Faz and CERF for their dashboards. Just simpler to keep a single way to update data and store everything in either CSV or Sheets I think. Would not be hard to recreate the Drive functionality if necessary in the future, and it wasn't streamlined before actually so I would do it differently (more like I have with Google Sheets) so it's a single line to read or update the files.

PR #16, opened 2023-08-01
Explicitly specify column typing

caldwellst:
Simple PR. Want to explicitly specify all the column typing when we load in Google Sheets for the system. Specifying them all in one location in case we load files in separate locations.


zackarno:
I like it! good to go

PR #17, opened 2023-08-01
Improve comms

caldwellst:
Will rebase after the `explicit-loading` branch has been pulled into `main`. This improves the IPC plot and makes a small change to the email text. Just review the `plot_ipc.R` and `email.Rmd` files. Putting as draft now and will change to full PR once rebased, but you can review whenever you want.


caldwellst:
Alright, this is ready for review! I fixed some of the IPC and other alert NLP. Basically making sure it doesn't appear in the email if it's NA. I fixed the IPC plot. And made a few other changes to the text that should be simple! Let me know if any other questions.

caldwellst:
Here's an example of the `plot_ipc()` output, FYI, so you don't have to manually re-run:

image

src/plots/plot_ipc.R
zackarno, line 1: i see this file function get's sourced in `src/plots/plot.R` with:

```
walk(
  .x = file.path(
    "src",
    "plots",
    c("plot_idmc.R", "plot_ipc.R", "plot_cholera.R")
  ),
  .f = source
)
```

but I don't get where the iso3 & country arguments are determined. If I just run `plot_ipc()` to test the func I get "Error: argument "iso3" is missing, with no default".

it's a bit hard to follow some of the processes going here as well as test the code. A couple things that make it confusing for me are:

- files are named the same. I have sever `run.R` tabs open in rstudio -- this seems ripe for confusion and potential errors
- not a clear order to run the processes (for testing & understanding) - I get that it doesn't matter in the GHA
- object are created and passed about through operations in a way that isn't very transparent/explicit (i.e I source a script which creates objects which are then passed to another script that get's sourced without being clear what they are).

zackarno:
The new plot is beautiful! This PR is taking a while to review since I haven't really done a "review" so far on this repo. Therefore, I am not going to attempt to review the whole repo now, but will relay some general relevant questions/comments that arise as I try to review this.

src/indicators/ipc/run.R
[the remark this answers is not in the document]
caldwellst: 1. Long term definitely don't want `get_country_names()`, but is there to help the CERF team who had specific names they wanted for their dashboard. Otherwise yes would be using `{countrycode}`.

2. `max(date_of_analysis)` filtration above happens when the date is grouped to the analysis periods and starts because the CH data needs de-duplication, as you note might be occurring. The slicing is simply dropping the last row because there is no value for the difference with the `lag()` done in lines 88 to 90 and 120 to 122, these values are just `NA` for the first observation in each group.

3. Good point, I think should be inverted!

4. Oh nice catch, came in during the new code changes. Thanks for this!

src/plots/plot_ipc.R
caldwellst: 1. So, on the `plot_ipc()` and other functions, I think you are overthinking what they are. They are just functions. Sourcing them is the base R way of making those functions available for use, which is what the lines you're quoting above do, nothing else. They don't rely on any external data except the explicit arguments you pass into it like most functions. The idea with moving to `{box}` is that these are then called in package format rather than sourcing the scripts, which makes R more like Python in its modular form, but the structure would still remain the same with utilities functions in `utils` and the plotting functions in `plots`, just referenced separately. So, once sourced, they are used in `.Rmd` files simply like `plot_ipc("NGA", "Nigeria")`, which works directly in the plot file if you want to test it without loading anything else.

2. Agree with you here, sourcing is a problem, but the way source is used here is simply to load the functional utilities in, nothing else. None of the `run.R` scripts for instance rely on each other. They are completely standalone and can be run on their own. I think this is less clear because of the terrible syntax of base R `source()`, and will be more clear when `{box}` is being used. As well, I think `{renv}`will be very useful here, as rather than using `source()` we may be able to use `renv::run()` which will run scripts in their own separate environment. I think there are a couple of instances where like email credentials are set in separate scripts, and those variables used elsewhere, so should again clean that up with `{box}`.

3. So, on the above, agree that we should be clear here and properly order the files and ensure the order is clear. I think this entails functionalizing code that is repeated and can be cleaned in that manner, which I am working to do in both of the other PRs (for generating AI summaries and for doing vector flagging). It can also be about ensuring documentation is clear. Definitely can see how it gets even more confusing when we look at the `.Rmd` files which are designed to be functionalized essentially. I don't think I agree about functionalizing the entire process so its input and output is consistent each time, but we can discuss and figure out the best way to simplify and clean up. For the purposes of this PR, this is completely out of scope however, so will create a separate issue with our discussion here.

4. What I have done in this PR is change the name of the `run.R` files, which I think was a very good idea.

In general, for this PR, I would look at it as each `run.R` for an indicator should be completely standalone, and output a wrangled file, a raw file, and a flags file. Then the main run pulls those together and generates emails as necessary, that's it. I'm around for any more questions and keep thinking about how it's confusing and how we might improve that so we can resolve that after these PRs in a larger PR that will move to using `{box}`, `renv::run()`, and generally clean up the flow for better troubleshooting and readability.

zackarno:
Great yeah a lot of that was out of scope for this PR, but became somewhat relevant when I was trying to test the functionality of some of the code contained in the repo generally as well as this PR. Anyways makes sense to continue the interesting discussion in the new issue you opened as well as on our call tomorrow.

From my side, the PR is good to merge! nice work, and I will get to the additional PRs ASAP.

PR #20, opened 2023-08-08
Upgrade to GPT4 model for summarization and improve long input handling

caldwellst:
Upgrades the AI summarisation to use GPT4, which has the biggest benefit of handling 8x the length of inputs of the model currently being used. Have also put into utilities the summarisation function. Will move to `box` once all these open PRs are done.


zackarno:
I have not yet dove into the `{openai}` package enough to understand these changes, but LGTM for merging. Perhaps we can chat directly about some of what is going here soon when we have some time.

PR #33, opened 2023-10-03
Improve emails

caldwellst:
Set up some various improvements to emails based on Sarah's comments. All changes detailed in `NEWS.md`, which I'm also implementing. As I have run from today the emails with these changes (and will continue to run from this branch instead of main until merged), the date is listed as October 3 so that it can be compared with the dates of emails being sent if ever necessary.

### DIFF
--- a/src/utils/flagging.R
+++ b/src/utils/flagging.R
@@ -22,6 +22,14 @@ box::use(lubridate)
 #' @param first_since `numeric` vector indicating when to flag the first positive
 #'     values. For instance, `180` would indicate flagging whenever there is a
 #'     positive value after 180 days.
+#' @param first_since_minimums `numeric` vector of minimum values such that if `x`
+#'     is above it, no flags are generated. If a single value, the value is
+#'     recycled. If the same length as `first_since`, the thresholds are matched
+#'     one-to-one with `first_since`. Since `first_since` is concerned with the
+#'     first value in a period of time, it only triggers on the first instance.
+#'     The minimum is therefore not just on the single first day, but is measured
+#'     against the following 60 days (which matches the period in which an email
+#'     would be generated, older flags would not generate any email).
 #' @param periods `numeric` vector of cumulation periods (in days) that flags
 #'     are calculated, used
 #'     for all of the `thresholds_...` arguments. For instance, `c(7, 30, 90)`
@@ -48,6 +56,7 @@ calculate_flags <- function(
     .data,
     x_col,
     first_since = NULL,
+    first_since_minimums = NULL,
     periods = NULL,
     thresholds_pcts = NULL,
     thresholds_static = NULL,
@@ -130,7 +139,7 @@ calculate_flags <- function(
     }
 
     if (is.null(thresholds_minimums)) {
-      thresholds_minimums <- 0
+      thresholds_minimums <- -Inf
     } else {
       ln_m <- length(thresholds_minimums)
       if (ln_p != ln_m && (ln_m != 1)) {
@@ -155,6 +164,32 @@ calculate_flags <- function(
     }
   }
 
+  # also check the same for first since
+
+  if (!null_fs) {
+    fs_order <- order(first_since, decreasing = TRUE)
+    first_since <- first_since[fs_order]
+    if (is.null(first_since_minimums)) {
+      first_since_minimums <- -Inf
+    } else {
+      ln_fsm <- length(first_since_minimums)
+      ln_fs <- length(first_since)
+      if (ln_fs != ln_fsm && (ln_fsm != 1)) {
+        stop(
+          "`first_since_minimum` must have the same length ",
+          "as `first_since` if passed to `generate_alerts()`, unless length 1, in ",
+          " which case it is recycled.",
+          call. = FALSE
+        )
+      }
+
+      # ensuring order is correct
+      if (ln_fsm > 1) {
+        first_since_minimums <- first_since_minimums[fs_order]
+      }
+    }
+  }
+
   # create the rolling sum columns necessary for flagging
   rs_periods <- c(first_since, periods)
   for (period in rs_periods) {
@@ -173,13 +208,32 @@ calculate_flags <- function(
   # check when we had a day with value above x
   # and 0 in the previous # number of dates
   if (!null_fs) {
-    first_since <- sort(first_since, decreasing = FALSE)
-    for (fs_period in first_since) {
-      .data <- dplyr$mutate(
-        .data,
-        "flag_first_{{ fs_period }}" := .data[[x_col]] > 0 & dplyr$lag(.data[[paste0("rs_", fs_period)]]) == 0
+    # create the forward looking 60 day period for checking the minimums
+    .data <- dplyr$mutate(
+      .data,
+      rs_next_60 = zoo$rollsum(
+        x = .data[[x_col]],
+        k = 60,
+        fill = "extend",
+        align = "left"
       )
-    }
+    )
+
+    df_fs <- purrr$map2(
+      .x = first_since,
+      .y = first_since_minimums,
+      .f = \(fs_period, fs_minimum) {
+        dplyr$mutate(
+          .data,
+          "flag_first_{{ fs_period }}" := .data[[x_col]] > 0 & dplyr$lag(.data[[paste0("rs_", fs_period)]]) == 0 & rs_next_60 >= fs_minimum
+        ) |>
+          dplyr$ungroup() |>
+          dplyr$select(paste0("flag_first_", fs_period))
+      }
+    ) |>
+      dplyr$bind_cols()
+
+    .data <- dplyr$bind_cols(.data, df_fs)
   }
 
   # if period is passed as an argument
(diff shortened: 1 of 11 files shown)


zackarno:
Nice work!

I like the updates and especially the new `news.md` file which is super helpful. 

I actually had just flagged some email wording to you in Slack that I think this PR probably fixes!

Some initial things i wanted to double check:

1. Now we that have changed the IDMC message to not include an `end_date` I am curious what happens if say in week 1 we flag a longer timeline trigger like a 90d anomaly and then in week 2 we flag  on a 7 day anomaly. I think it will just report the number of cases since the start of the 90 days? I think this makes sense, but I feel like we should also try to highlight the more recent increase in the last 7 days?
2. I just want to make sure that previous records/messages in the previous file that we load don't get changed? I think they only have because of project changes/developments, but just want to make sure.

caldwellst:
> Now we that have changed the IDMC message to not include an end_date I am curious what happens if say in week 1 we flag a longer timeline trigger like a 90d anomaly and then in week 2 we flag on a 7 day anomaly. I think it will just report the number of cases since the start of the 90 days? I think this makes sense, but I feel like we should also try to highlight the more recent increase in the last 7 days?

So yeah in this situation we are doing as you say, although the start date would be when the displacement starts. In the instance where there are two alerts at the same time, we would take the latest one anyway for flagging, but all would be highlighted in the email.

> I just want to make sure that previously logged records/messages in the "previous" don't get changed? I think they only have because of project changes/developments, but just want to make sure.

Yeah they have changed, but we have `df_emailed` which is tracking every message that is emailed. So we can always look back historically at what we've sent out.

zackarno:
when using `{box}` can we adjust the project-specific settings to use the native pipe `|>` as i think it will make debugging easier?

zackarno:
why not just `pmap_dfc()`

https://github.com/OCHA-DAP/pa-global-monitoring/blob/f5814ee68ed9a0a22ffb7462755178aa84d0f2a0/src/utils/flagging.R#L243

src/utils/flagging.R
```diff
@@ -173,13 +208,32 @@ calculate_flags <- function(
...
+          "flag_first_{{ fs_period }}" := .data[[x_col]] > 0 & dplyr$lag(.data[[paste0("rs_", fs_period)]]) == 0 & rs_next_60 >= fs_minimum
+        ) |>
+          dplyr$ungroup() |>
+          dplyr$select(paste0("flag_first_", fs_period))
+      }
+    ) |>
+      dplyr$bind_cols()
+
+    .data <- dplyr$bind_cols(.data, df_fs)
   }
```
zackarno, line 237: what do you think about using  `purrr$map2_dfc()` & leaving `iso3` and `date` so that we can the `left_join(.data,df_fs, by =c("iso3","date"))` -- this might provide some guard rails against anything strange happening?

src/utils/flagging.R
```diff
@@ -22,6 +22,14 @@ box::use(lubridate)
...
 #'     values. For instance, `180` would indicate flagging whenever there is a
 #'     positive value after 180 days.
+#' @param first_since_minimums `numeric` vector of minimum values such that if `x`
+#'     is above it, no flags are generated. If a single value, the value is
+#'     recycled. If the same length as `first_since`, the thresholds are matched
+#'     one-to-one with `first_since`. Since `first_since` is concerned with the
+#'     first value in a period of time, it only triggers on the first instance.
+#'     The minimum is therefore not just on the single first day, but is measured
+#'     against the following 60 days (which matches the period in which an email
+#'     would be generated, older flags would not generate any email).
```
zackarno, line 32: how does 60 match the period in which an email would be generated?

src/utils/flagging.R
```diff
@@ -173,13 +208,32 @@ calculate_flags <- function(
...
-    }
+    )
+
+    df_fs <- purrr$map2(
+      .x = first_since,
+      .y = first_since_minimums,
+      .f = \(fs_period, fs_minimum) {
+        dplyr$mutate(
+          .data,
+          "flag_first_{{ fs_period }}" := .data[[x_col]] > 0 & dplyr$lag(.data[[paste0("rs_", fs_period)]]) == 0 & rs_next_60 >= fs_minimum
```
zackarno, line 228: maybe a couple comments here - took a while for me to figure out what was going on

caldwellst:
> why not just `pmap_dfc()`
> 
> https://github.com/OCHA-DAP/pa-global-monitoring/blob/f5814ee68ed9a0a22ffb7462755178aa84d0f2a0/src/utils/flagging.R#L243
> 
> For the below: same comment as [#33 (comment)](https://github.com/OCHA-DAP/pa-global-monitoring/pull/33#discussion_r1373569864)
> 
> https://github.com/OCHA-DAP/pa-global-monitoring/blob/f5814ee68ed9a0a22ffb7462755178aa84d0f2a0/src/utils/flagging.R#L261-L268



> why not just `pmap_dfc()`
> 
> https://github.com/OCHA-DAP/pa-global-monitoring/blob/f5814ee68ed9a0a22ffb7462755178aa84d0f2a0/src/utils/flagging.R#L243
> 
> For the below: same comment as [#33 (comment)](https://github.com/OCHA-DAP/pa-global-monitoring/pull/33#discussion_r1373569864)
> 
> https://github.com/OCHA-DAP/pa-global-monitoring/blob/f5814ee68ed9a0a22ffb7462755178aa84d0f2a0/src/utils/flagging.R#L261-L268

`pmap_dfc()` and all variants are no longer recommended and have been deprecated by `{purrr}` in lieu of explicit `map()`, `walk()`, `flatten()`, etc. and then `list_rbind()` or `list_cbind()`.

zackarno:
> `pmap_dfc()` and all variants are no longer recommended and have been superseded by `{purrr}` in lieu of explicit `map()`, `walk()`, `flatten()`, etc. and then `list_rbind()` or `list_cbind()`.

oh interesting i had no idea! is `list_rbind()` preferred of `bind_rows()` as well?

src/utils/flagging.R
caldwellst: Thanks, good point. I have added a comment to clarify why I did it this way, which is essentially, we ensure that you don't have to have `iso3` or `date` columns or ANY specific columns, and if you have additional columns they won't be lost. We simply add only the new columns.

caldwellst: Which I think makes it robust to strange behavior in case you passed something in with extra columns that might get lost for instance.

caldwellst:
> when using `{box}` can we adjust the project-specific settings to use the native pipe `|>` as i think it will make debugging easier?

Yes, good idea, make it explicit for that project!

caldwellst:
> > `pmap_dfc()` and all variants are no longer recommended and have been superseded by `{purrr}` in lieu of explicit `map()`, `walk()`, `flatten()`, etc. and then `list_rbind()` or `list_cbind()`.
> 
> oh interesting i had no idea! is `list_rbind()` preferred of `bind_rows()` as well?

Hmm, it seems not, cause I had to Google just in case. https://community.rstudio.com/t/purrr-list-rbind-versus-dplyr-bind-rows/164411 

I think it might be because `bind_rows()` is meant to be a tidy approach of the wide functionality of `rbind()`, whereas `list_rbind()` is simply for `{purrr}` functionality?

src/utils/flagging.R
caldwellst: Okay, so I think the confusion here and below definitely deserves some clarification. But the whole redesign basically came out of some issues with the emails, mainly unique to IDMC, where delays in reporting can generate older flags.

So, for instance, data could be added for 30 days ago that would generate a flag on the first displacement since a year.

Let's say 400 people in Mali were displaced on October 1. However, maybe there was also displacement last week, 300 people in Mali on October 20. And the data was added today, the 27th of October. 

Previously, the flag would just look at displacement from that first alert. But, that meant the emails messaging would maybe miss reporting on all the displaced values since then. It would say 400 displaced on October 1, ignoring the latter data. So, we wanted to instead message about all of the available data SINCE the flag was generated.

So that was what `rs_next_60` is. Basically, emails are not generated if the alerts are older than 60 days, so with this sum we can assume we will have all latest displacement figures in our message. In this case, 700 people have been displaced SINCE October 1 (and a plot to contextualise all this).

Now, we also got feedback to not flag if displacement is under a certain amount for these first since flags. But, at what level do we look at that? In this case, the threshold is 500, so if we just look at October 1, we wouldn't flag. But if we look across the potential period for flagging, we have 700, so we would flag. So what I did was look across the whole next 60 days and see if that meets that threshold.

So basically we have system that needs to look back a period of time to check that nobody was displaced, and over the next period of time to make sure that a certain amount of people have been displaced. Maybe we could only look backwards, but what we are doing now is looking from the date of alert, when we would list the start date of the alert, and in our messaging, say that displacement started. So there is definitely value there. And hard to know how to easily look back and say in the past 60 days, we see X displacement, but past 180 days + 60 we see none. Might not be that much simpler.

But, putting this in an open issue so we can discuss further!

caldwellst:
Okay, I have created an issue #34 so we can further discuss the approach to flagging in the utilities, where we can clean it up/break up functions for better clarity and potentially change the approach. Let's continue the conversation there but will merge this for now as that would be another very large PR.

PR #38, opened 2023-12-11
Fix start_dates for cholera

### LINKED ISSUE #37: start_date parsing + data quality
More of a data quality issue that i think we should just flag up probably. But documenting it here:

I was looking at the cholera data and noticed that all `start_date` for `week_date`  `2023-09-25` was not properly parsing. I noticed this is because the `start_of_reporting_period` format changed. Should we just flag this to the team doing the scraping?

Digging a bit more into the `start_of_reporting_period` I do see that the formatting does change quite a bit over time (either `dmy` or `mdy`), therefore the code here either fails to parse some records  or in some cases likely parses incorrectly.

https://github.com/OCHA-DAP/pa-global-monitoring/blob/2c05b5bd038bf72e664689234a88c89583773c11/src/indicators/cholera/update_cholera.R#L50-L55

there would be a light cleaning step to where we could adjust the above w/ a conditional `dmy()` or `mdy()` based on the values (i.e months can't be >12) ... but this wouldn't ensure they were all correct.

caldwellst:
Closes #37. Checks for 3 alphabetic characters in the date, and if they are present, parses using `dmy()`. If the date was prior to September 25, 2023, the parsing is still done as `dmy()` (since this is how those files have always been parsed. Then, otherwise, it parses as `mdy()`, which is the new format used by the WHO AFRO bulletins.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,10 @@
+## December 11, 2023
+
+### Indicator: cholera
+
+- Adjusted start date parsing adjusted to address change in date format from
+`%d-%b-%y` to `%m/%d/%Y` in the WHO AFRO bulletins.
+
 ## October 3, 2023
 
 ### Email alerts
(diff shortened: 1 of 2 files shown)


zackarno:
nicely done - great fix! good to merge :-)

PR #39, opened 2023-12-13
Data backend move to GCS

caldwellst:
Move the system from Google Drive and Sheets to GCP! Have run all tests in GHA and done todays monitoring run, all working.


src/indicators/cholera/README.md
```diff
@@ -35,9 +35,9 @@ date of the outbreak. The end date is the date that the flag was raised.
 
 ## Output datasets
  
-`flags_cholera.gdsheets`: flagging dataset.
+`flags.parquet`: flagging dataset.
 
-`raw_cholera.gdsheets`: raw cholera output from the CERF scraping.
+`raw.parquet`: raw cholera output from the CERF scraping.
```
zackarno, line 40: `raw.parquet`: raw cholera output from CERF scraping (downloaded as .csv and converted to .parquet)

src/indicators/idmc/update_idmc.R
```diff
@@ -20,7 +20,7 @@ box::use(../../utils/flagging)
 ##############################
```
zackarno, line 20: can't comment on unchanged code, so putting comment for code line 11 here:

```
# first set the root search path for utilities
box::use(gs = ../../utils/google_sheets)
#> Error in box::use(gs = ../../utils/google_sheets) : 
#>   unable to load module “../../utils/google_sheets”; not found in “/Users/zackarno/Documents/CHD/repos/pa-#> #> global-monitoring/src/indicators/idmc”
```

src/indicators/cholera/update_cholera.R
```diff
@@ -172,17 +172,17 @@ df_cholera_flags_final <- dplyr$semi_join(
 #### SAVE IPC  DATA ####
 ########################
 
-gs$update_gs_file(
+cs$update_gcs_file(
```
zackarno, line 175: need to add to this script
```
box::use(cs = ../../utils/cloud_storage)
```

src/indicators/idmc/update_idmc.R
```diff
@@ -88,7 +88,7 @@ df_idmc_flags <- flagging$generate_alerts(
 #### COMPARE WITH EXISTING FILES ####
 #####################################
 
-df_idmc_flags_prev <- gs$read_gs_file("flags_idmc") |>
+df_idmc_flags_prev <- cs$read_gcs_file("output/flags/idmc.parquet") |>
```
zackarno, line 91: need to add
```
box::use(cs = ../../utils/cloud_storage)
```

think you want to get rid of  google_sheets call in all scripts (here & cholera)

```
box::use(gs = ../../utils/google_sheets)
```

zackarno, line 91: getting this error on local machine... weird because other `cs` call to load parquets are working

prevents creation of this obj and subsequent
```
df_idmc_flags_prev <- cs$read_gcs_file("output/flags/idmc.parquet") |>
  dplyr$mutate(
    email = FALSE
  )
#> ℹ 2023-12-15 11:40:18.308348 > Request Status Code:  404
#> Error in `abort_http()`:gs/idmc.parquet
#>  ! http_404 Unspecified error
#> Run `rlang::last_trace()` to see where the error occurred.
#> ✖ Downloading output/flags/idmc.parquet ... failed

```

src/utils/cloud_storage.R
```diff
@@ -0,0 +1,181 @@
...
+  path = Sys.getenv("GLOBAL_MONITORING_JSON")
+)
+
+googledrive$drive_auth(
+  path = Sys.getenv("GLOBAL_MONITORING_JSON")
+)
+
+gcs$gcs_auth(
+  json_file = Sys.getenv("GLOBAL_MONITORING_JSON")
+)
```
zackarno, line 159: maybe a separate file for these authorizations to make it easier for newcomers to navigate file structure (improve long term maintenance)?

Could also separate deprecated gdrive functions (same purpose i think)

src/utils/cloud_storage.R
```diff
@@ -0,0 +1,181 @@
...
+  )
+}
+
+#' Parse parquet file
+#'
+#' Helper function to parse parquet file.
+#'
+#' @param object Object to parse
+gcs_parse_parquet <- function(object) {
+  con <- gzcon(rawConnection(httr::content(object)))
```
zackarno, line 41: never seen this! is it required to read in parquet from GCS?

zackarno (commented):
Awesome - love the move to Google Cloud Storage Buckets! I think this will be way more robust as we transition our data storage strategies.

Left a couple comments below

if GHA is running the script without the above changes integrated (missing `box::use(cs = ../../utils/cloud_storage)` & error with `cs$read_gcs_file("output/flags/idmc.parquet")` ) , does that mean it is silently skipping/creating issues? If so what guardrails can we put up?

src/utils/cloud_storage.R
caldwellst: Yeah, you could either separately download the file without any parsing, and then just read in. Or you can specifically create a parsing function that you just pass in to the function that reads from the connection.

caldwellst:
Nothing to do with `{box}` in the end. What was happening is apparently if you use `find -exec` in bash to loop through files and run them, as happens in the `global-monitoring.yaml` file, [errors generated from the loops are just passed over](https://apple.stackexchange.com/questions/49042/how-do-i-make-find-fail-if-exec-fails)! So I've had to switch to globbing.

caldwellst:
Beyond that, all your comments should now be addressed! Thanks for the review, super helpful!

src/utils/cloud_storage.R
zackarno: I wonder if we can use the approach outlined here which allow  you to read the parquet files in directly?

https://arrow.apache.org/docs/10.0/r/articles/fs.html

I couldn't get it to work in the global-monitoring bucket, but i am wondering if that has something to do with permissions settings? because the examples in the article do work

zackarno:
Sweet, good job figuring out that issue in the yaml!!! LGTM, but check out my question here: https://github.com/OCHA-DAP/pa-global-monitoring/pull/39#discussion_r1431925894 would be curious to hear what you think on that point

PR #41, opened 2024-02-11
Mailchimp implementation

caldwellst:
Test Mailchimp implementation. There is a lot of module level documentation, but not overarching updates to the README files yet. However, all of the functionality is tested in `update_cholera.R` if you want to explore all of the different code. There are still some issues to iron out with Mailchimp, such as how to get conditional content to appear in the archive. However, overall it's working!

The priority when I'm back would be to make minor fixes to get the actual emails looking nicer, ensuring archive URLs work (may just need to create a campaign without conditional blocks and have that as the public URL). I also need to do the same code for IPC and IDMC.

However, a review would be very valuable just so you're familiar with the code and can start helping me improve the robustness of it. The other priority beyond it working is to have it using available orchestration tools and better techniques to ensure that each piece is isolated and robust to failure.


caldwellst:
@zackarno @PaulineNimo @t-downing  In case you're interested!

hannahker:
@caldwellst what's your thinking on merging in this branch? Would be nice to work against `main` for other features!

hannahker:
Or to @zackarno's idea, we could maybe rename this to a `dev` branch that we're all working off of (and make it the default branch)?

caldwellst:
Yes, can do now, couldn't merge to main earlier because you can't run CRON jobs off of main and the old system was still running for CERF. However, we can now shift everything over.

PR #45, opened 2024-03-18
Azure storage

caldwellst:
Shift the system over to using Microsoft Azure Data Storage.


caldwellst:
Note that I have run a test run action #333 and a full action run #334 both of which succeeded in accessing and saving to the file storage.

src/indicators/idmc/update_idmc.R
```diff
@@ -114,7 +114,7 @@ df_idmc_flags_new <- dplyr$anti_join(
   dplyr$group_by(iso3) |>
   dplyr$mutate(
     email = ifelse(
-      Sys.Date() - end_date > 60 | start_date != max(start_date), # do not email for old events updated or if not the latest alert
+      Sys.Date() - end_date > 60 | start_date != max(start_date, -Inf), # do not email for old events updated or if not the latest alert
```
zackarno, line 117: whats the addition of  `-Inf` doing here? I'm sure it's fine, just thought easier if you explain

renv/activate.R
```diff
@@ -31,6 +31,14 @@ local({
     if (!is.null(override))
       return(override)
 
+    # if we're being run in a context where R_LIBS is already set,
```
zackarno, line 34: I don't really understand why when run `renv::retore()` this file is updated on my end to exactly what is here and then git flags that i have uncomitted changes. Not sure if you observed and or understand this either?

src/utils/cloud_storage.R
```diff
@@ -1,91 +1,138 @@
...
+)
 
-# this bucket used to read and write from the GCS parquet
-bucket <- arrow$gs_bucket(
-  bucket = "global-monitoring",
-  json_credentials = Sys.getenv("GLOBAL_MONITORING_JSON")
+# blob object for HDX Signals, used to read and write data
+blob <- az$blob_container(
+  endpoint = blob_endpoint,
+  name = "hdx-signals"
```
zackarno, line 137: For me, the workflow would be easier to follow if these two calls were just added at the top of each `update_*.R` scripts or perhaps loaded in a different way as this is script is all functions other than these last calls. I feel like running/sourcing a module/or script in an analysis script that actually loads/assigns objects in the the env feels a bit weird for R-users, but might be more common in other languages/production related work flows?

zackarno (approved):

src/indicators/idmc/update_idmc.R
caldwellst: Removing some of those warnings that you were seeing in the GitHub actions. Basically if you have a data frame with no rows (sometimes no new flags), then a warning is generated from `max(start_date)` because it's a vector of `0` length. `-inf` just simply removes that warning so that `max()` returns some value at all times.

renv/activate.R
caldwellst: Not entirely sure. Reading the documentation, we have this description:

> Restore a project's dependencies from a lockfile, as previously generated by [snapshot()](http://127.0.0.1:27111/help/library/renv/help/snapshot). renv::restore() compares packages recorded in the lockfile to the packages installed in the project library. Where there are differences it resolves them by installing the lockfile-recorded package into the project library. If clean = TRUE, restore() will additionally delete any packages in the project library that don't appear in the lockfile.

Maybe you just need to run `clean = TRUE` to ensure your local differences (extra packages) are removed.

src/utils/cloud_storage.R
caldwellst: Remember we are using `box::use()`. If we were using `source()`, these objects would be loaded into the environment. However, as is, this is just standard programming practice in a way that I am defining global objects that can be used across the different functions in this specific module. So users never use or even see these objects, but they are used within the `update_az_file()` and `read_az_file()` functions, which are called frequently in other modules.

So, I think adding this elsewhere would violate the do not repeat yourself rule.

zackarno: fair, I don't really have any issue with it as i'm familiar with the code. I was more thinking if someone new was troubleshooting down the road.. how would they know these objects are loaded and made here specifically? But yeah im sure your right and im just not used to seeing it like this.

caldwellst: Yeah, but in this case, those objects, for instance `blob` are defined in that module `cloud_storage.R`, but they are never used outside of it! Only the exported functions are, which are clearly imported in other scripts, so I would hope it's clear to the user. I think I agree with your point about making sure anything `source()` is just functions, but `box::use()` ONLY imports things that have the `@export` roxygen tag, so takes care of this nicely I think!

PR #71, opened 2024-05-10
Ensure that we get full list of audience members

caldwellst:
Pulling in our list of audience members, noticed after our tests today only 10 members were receiving the alerts, turns out that the default return from the Mailchimp API was 10, so have set it up so it automatically pulls in 1,000 by default and paginates through if necessary.

I have previously used `httr2::req_perform_iterative()` with iterator functions, but was too tired to figure out how to do it this time and couldn't see any link follow on in the response object, just `total_items`, so thought this was the best way, but happy for you to suggest/improve!

### DIFF
--- a/src/email/mailchimp/audience.R
+++ b/src/email/mailchimp/audience.R
@@ -18,13 +18,28 @@ box::use(../../utils/country_codes)
 #'
 #' @export
 mc_members <- function() {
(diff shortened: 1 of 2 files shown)


hannahker (approved):
Good catch! Not sure how to test this, but the logic makes sense to me.

PR #78, opened 2024-05-13
Align email link color with hdx-sapphire color in visualization

caldwellst:
Changing color of links in email HTML file to match the color in our plots, `gghdx::hdx_hex("sapphire-hdx")`. You can review since you put in the restrictions on the repo, mwahaha.


hannahker (approved):
Nice one 😎

Issue #79, opened 2024-05-13
Enforce code styling


caldwellst:
To avoid me being really annoying in PRs, would be good to have code styling and checks using `pre-commit`. As @hannahker has noted, we can use `{lintr}` to lint the code and force manual edits to match the style prior to committing. We could even use `{styler}` to automatically adjust the code, but would need some manual edits to the system. For instance, when testing it was messing with `{box}` to change `box::use(a/b/c)` to `box::use(a / b / c)`.

I propose that we use the tidyverse style guide. https://style.tidyverse.org/index.html

hannahker:
Closed with #81

PR #81, opened 2024-05-14
Lint repo

caldwellst:
Set up linting for the repo and lint `main`. Based off discussion in #79 and the initial approach in #80.


.github/workflows/test-renv copy.yaml
hannahker, line 1: Update the file name here?

src/email/mailchimp/custom_segmentation.R
```diff
@@ -85,52 +85,66 @@ mc_subscriber_emails <- function(df_ind, iso3, test) {
...
-              member$email_address
-            )
-          }
-        }
-        return(
-          NA_character_
-        )
-      }
-    )
+    member_emails <- interest_emails(interest = df_ind$mc_interest, geo_ids = geo_ids, test = test)
```
hannahker, line 88: Nice 👌

.lintr
hannahker, line 1: Oh nice! Assuming this helps things work automatically with `styler`?

hannahker (approved):
On scan through this looks good to me! Do you think it's worth any further testing before merging? 

Just need to update the `yaml` file name though.

.lintr
caldwellst: Yeah! Well for `{lintr}`, not `{styler}`. Automatically sets our preferences. We have to change a few things so it means if we run it locally we will get the same results as in the cloud.

caldwellst:
Alright, merging! For now given the branch is passing the only test we have, the lint we implemented here, will merge. As we develop more and more tests, we can be more careful in our development but I'm prioritising speed at the moment hahaha.

Issue #88, opened 2024-05-15
where is un_geodata.geojson initially created?


zackarno:
where is `un_geodata.geojson` initially created? I can't seem to find it

hannahker:
@caldwellst is it [this file](https://geoportal.un.org/arcgis/home/item.html?id=d7caaff3ef4b4f7c82689b7c4694ad92)?

caldwellst:
Yes, that's it!

hannahker:
Hmm do you see a way that we could download this programmatically? I'm just seeing a manual download button

caldwellst:
No, that's why I just manually downloaded. Is there a way to access via web services @zackarno? I'm not too familiar with that kinda map data provision.

caldwellst:
This needs to be added to our documentation!

caldwellst:
Still needs to be documented all of these files, will put in `src-static`. Note also that I've added a new manually procured geojson with complex boundaries in #178.

PR #92, opened 2024-05-16
Fix static data updates

hannahker:
Minor updates to ensure that the static data update runs from the GitHub action with informative logging and status updates. 

Issues still to resolve: 

- [x] Connection errors when running action ([for example](https://github.com/OCHA-DAP/hdx-signals/actions/runs/9117039573/job/25067015146))
- [x] The action should fail when R errors out

If we're assuming that all files under `inputs/` in the storage container should be updated by [this workflow](https://github.com/OCHA-DAP/hdx-signals/blob/main/.github/workflows/update_assets.yaml):

- [x] `adm0/` 
- [x] `cities/` 
- [x] `country_info.parquet`
- [x] `hrp_countries.parquet` 
- [x] `iso3_map_settings.json`
- [x] `centroids/` -- Updates work locally but not in GH runner
- [x] `indicator_mapping.parquet`
- [x] `idmc_country_links.parquet` 
- [ ] `acled_info.parquet` -- Updates work locally but not in GH runner
- [ ] `un_geodata.geojson` -- Updated manually and used in the code. Can this be updated programmatically?
- [ ] ~`country_taxonomy_backup.parquet` -- Removing~
- [ ] ~`region_bbox.parquet` -- Removing~
- [ ] ~`region_bbox.geojson` -- Removing~


caldwellst:
> The action should fail when R errors out

This is an issue with using `find` in GitHub Actions. This is why in the original file we were using `globstar`. See this https://apple.stackexchange.com/questions/49042/how-do-i-make-find-fail-if-exec-fails

caldwellst:
> Connection errors when running action ([for example](https://github.com/OCHA-DAP/hdx-signals/actions/runs/9117039573/job/25067015146))

Does this work when running locally?

caldwellst:
`un_geodata.geojson`: See #88 
`idmc_country_links.parquet` and `indicator_mapping.parquet`. I put those together manually, just checking all country links on the IDMC website, and then piecing together `indicator_mapping.parquet` as I was developing. I think we can hardcode an update script that we adjust as necessary. A really nice and easy trick is that you can generate the code to create the data frame in R super easily. Just load it in, save to `df` and then use `dput(df)`. However, it's not that clean / easy to update, whereas `dplyr::tribble()` is a nice way to rowwise create data.

hannahker:
@caldwellst 

Aha nice find! 

> This is an issue with using find in GitHub Actions. This is why in the original file we were using globstar. See this https://apple.stackexchange.com/questions/49042/how-do-i-make-find-fail-if-exec-fails

It does! I'm thinking that this could be from the `{}` in the `DSCI_AZ_ENDPOINT` environment variable. I'm thinking we can just hard-code this in the code itself since this doesn't contain any sensitive data. 

> Does this work when running locally?

.github/workflows/update_assets.yaml
```diff
@@ -16,6 +16,8 @@ jobs:
         runs-on: ubuntu-latest
         env:
           GMAS_TEST_RUN: ${{ inputs.test || 'FALSE' }}
+          DSCI_AZ_SAS_DEV: ${{ secrets.DSCI_AZ_SAS_DEV }}
```
caldwellst, line 19: Oh, was `DSCI_AZ_ENDPOINT` also not set here?

hannahker: Hah, it was also missing 😆 But I was testing and it still seems to cause some issues even when I put it back. So might just leave as is...

hannahker:
@caldwellst this should be ready for final review now. As discussed, we'll come back to the ACLED data issues in future work. For now, I've left the bash command as using `find`, so that this script can run through all other update files, despite the ACLED data failing.

src-static/update_indicator_mapping.R
```diff
@@ -0,0 +1,32 @@
...
+)
+# nolint end
+
+fname <- "input/indicator_mapping.parquet"
+cs$update_az_file(
+  df = df,
+  name = fname
+)
+
+log_info(paste0("Successfully downloaded indicator mapping and saved to ", fname))
```
caldwellst, line 32: Ooh sorry, can you just put this as "created indicator mapping"

src-static/update_acled_info.R
```diff
@@ -4,46 +4,59 @@ box::use(dplyr)
...
-    acled_hdx_country = Country |>
-      stringr$str_remove_all("[^[:alnum:]\\s-\\.]") |>
-      stringr$str_remove_all("\\.$") |>
-      tolower() |>
-      stringr$str_replace_all("\\s|\\.", "-"),
-    acled_hdx_url = as.character(
-      glue$glue(
-        "https://data.humdata.org/dataset/{acled_hdx_country}-acled-conflict-data"
+# TODO: Remove once this passes from GH runner
+if (interactive()) {
```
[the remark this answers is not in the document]
hannahker: yeah, still failing :(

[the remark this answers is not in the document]
hannahker: Oh sorry, by that I mean that the ACLED access still doesn't work, but the overall action is successful with `interactive()`

PR #95, opened 2024-05-17
IPC food insecurity

caldwellst:
Code to implement the food insecurity signals that went out today. Includes changes to some of the utilities function to improve how we pass alerts data frame through `generate_signals(), generate summarisation's, and other stuff. Happy to take you through if necessary, although in general I need to give everything a good run through once we are up and running anyway!


hannahker (approved):
This looks good to me! I like that the structure of the indicators `utils/` is so standardized. Definitely picking back up on my R looking at this!

src/indicators/ipc_food_insecurity/utils/summary_food_insecurity.R
```diff
@@ -0,0 +1,197 @@
...
+      summary_long = purrr$pmap_chr(
+        .l = list(
+          url = link,
+          country = country,
+          ch = ch
+        ),
+        .f = ipc_ch_summarizer
+      ),
+      prompt_short = paste(
+        "Please condense the following information into a single 10 word line,",
```
hannahker, line 26: Agree with #97 that it would be excellent to keep these in `.txt` files

PR #96, opened 2024-05-18
Implement capacity to view images in locally generated HTML

### LINKED ISSUE #67: Find method to view local images in HTML preview
When developing indicator signals, it's useful to be able to test how the campaign will look, including the content and images. For this, we can run `generate_signals(..., test = TRUE)`. If doing this with `GMAS_TEST_RUN` as `FALSE`, all of the content is still generated and saved to Mailchimp and Azure, and campaigns can even be sent to Mailchimp subscribers tagged as `hdx-signals-test`.

However, if doing this with `GMAS_TEST_RUN` as `TRUE`, all of the templates and imagery are generated locally. The campaign preview is opened up in your default browser using a locally saved HTML file with local links to the images.

The issue is that modern browsers apparently don't actively link to local imagery, see for instance this [SO post](https://stackoverflow.com/questions/28194444/how-can-i-display-an-image-from-the-local-machine-on-a-webpage). Because of this, the images do not appear in the preview. We can still individually open up the local links from the file, or set `GMAS_TEST_RUN` to `FALSE` and test through Mailchimp.

It would be ideal and faster, however, if we could quickly create local previews. So would be good to find a way to make the local imagery appear in the preview. Maybe there is a specific browser for this we could specify as default in `utils$browseURL()`?

caldwellst:
Closes #67


hannahker (approved):
This looks good to me!

PR #100, opened 2024-05-20
WFP market monitor

caldwellst:
Initial WFP market monitor run. Putting in draft format because we are awaiting changes to the input dataset which might allow additional plots or actual summarisation for our emails.


caldwellst:
Data has been updated by WFP to include what we need. Still extremely limited in content, but this may be all we have available, so we can at least review what is here. Have merged to main and incorporated the new changes from @hannahker  like logging and actions YAML file.

zackarno:
This PR is still a WIP, right?

caldwellst:
Ready for review now. We will do separate updates if they provide new data.

src/utils/cloud_storage.R
```diff
@@ -20,24 +20,25 @@ hs_logger$configure_logger()
...
 #'     and file extension, such as `.parquet`.
-#' @param stage Stage in the Azure storage to read from, either `prod` or `dev`.
+#' @param blob Blob in the Azure storage to read from, either `prod`, `dev`, or `wfp`.
 #'
 #' @returns Data frame.
 #'
 #' @export
-read_az_file <- function(name, stage = c("prod", "dev")) {
-  blob <- stage_to_blob(stage)
+read_az_file <- function(name, blob = c("prod", "dev", "wfp")) {
```
hannahker, line 40: Nice! This is clearer now

src/utils/cloud_storage.R
```diff
@@ -142,20 +143,21 @@ az_file_detect <- function(pattern = NULL, stage = c("prod", "dev")) {
...
-stage_to_blob <- function(stage = c("prod", "dev")) {
-  stage <- rlang$arg_match(stage)
-  if (stage == "prod") {
-    blob_prod
-  } else {
-    blob_dev
-  }
+get_blob <- function(blob = c("prod", "dev", "wfp")) {
+  blob <- rlang$arg_match(blob)
+  switch(
```
hannahker, line 155: Love this syntax 😍

src/utils/cloud_storage.R
```diff
@@ -198,3 +201,9 @@ blob_prod <- az$blob_container(
   endpoint = blob_endpoint_prod,
   name = "hdx-signals-mc"
 )
+
+# blob object for WFP, where the market monitoring data is stored
+blob_wfp <- az$blob_container(
+  endpoint = blob_endpoint_dev,
```
hannahker, line 207: Is there a reason we wouldn't put this in the prod blob container? It'd be great to keep things centralized there.

src/indicators/wfp_market_monitor/utils/raw_market_monitor.R
```diff
@@ -0,0 +1,13 @@
+box::use(cs = ../../../../src/utils/cloud_storage)
+
+#' Download raw market monitoring data
+#'
+#' Downloads raw market monitoring data from Azure database. Is uploaded directly
+#'
+#' @export
+raw <- function() {
```
hannahker, line 8: Could you clarify how the overall monitoring process will work for this data since we're referencing static inputs? How often will WFP send new data? Do we want to have validation that any new CSV files have the same structure as expected?

src/utils/cloud_storage.R
caldwellst: Yes, @isatotun can hopefully get this into prod but previously stuff she was working on was put into `dev` due to access issues she was having that needed solutions from Saeid. Will create a separate issue because we want to shift everything she's using from `dev` to `prod`.

src/indicators/wfp_market_monitor/utils/raw_market_monitor.R
caldwellst: Yes, good idea to have the test! They are updating whenever they have updates, currently maybe every other week, but need to get confirmation of them that this pipeline is going to be frequently updated and monitored. THye are by far the most difficult data provider.

caldwellst: Currently just checking that at least the columns we need exist.

caldwellst:
Merged with your new functions to stop checks for the env vars, now looks much better all around! Ready for another review @hannahker .

src/utils/cloud_storage.R
hannahker: Copy that 🫡

hannahker (approved):
Nice updates! This looks good to merge! 💃 

We'll just have to make sure to clearly document that this indicator is a bit different since the dataset is updated manually.

PR #102, opened 2024-05-20
Backoff req

caldwellst:
Given the issues with Mailchimp servers today, implemented some simple retry logic into the API requests. Also made a quick change in ACLED plots as well because was trying to run ACLED with the retry logic during the downtime.


hannahker (approved):
LGTM 💃

PR #103, opened 2024-05-20
Fix test segmentation

caldwellst:
So, it turns out that updating static segments with a list of emails was not REMOVING emails already on the list. So to fix that, first need to remove the emails to make it empty, then add back the emails we want to send to.

### DIFF
--- a/src/email/mailchimp/custom_segmentation.R
+++ b/src/email/mailchimp/custom_segmentation.R
@@ -44,6 +44,14 @@ mc_email_segment <- function(indicator_id, iso3, test = FALSE) {
     test = test
   )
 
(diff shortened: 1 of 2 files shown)


src/email/mailchimp/custom_segmentation.R
```diff
@@ -44,6 +44,14 @@ mc_email_segment <- function(indicator_id, iso3, test = FALSE) {
     test = test
   )
 
+  # first we empty the segment before updating it
+  segments$mc_update_static_segment(
```
hannahker, line 48: Nice one 🌟

hannahker (approved):

PR #106, opened 2024-05-20
Add output for short summary function

caldwellst:
Quick fix, no output was being given in the case the `if()` state wasn't evaluated.


hannahker (approved):
Looks like a good catch 🤓

PR #107, opened 2024-05-21
GPT-4o

caldwellst:
Move to GPT-4o, which is the new flagship from OpenAI and also half the price as GPT-4 turbo. Also took in a couple of other edits: request to try a new title change, and removed `triage_signals()` from the update scripts.


src/indicators/who_cholera/update_cholera.R
hannahker, line 1: Is it worth just removing this indicator altogether if it's not going to be used?

hannahker (approved):
Looks good!

src/indicators/who_cholera/update_cholera.R
caldwellst: Yes probably...

PR #108, opened 2024-05-21
Remove warnings

caldwellst:
Remove warnings from update scripts, or at least some of them. Adds default values to `min()` and `max()` calls that shouldn't interrupt processing (basically, if they are used in `filter()` and now won't generate warnings). Returns empty data frame immediately if the alerts data frame is empty, which should avoid majority of warnings.


hannahker:
Oh, amazing! Was just wrestling with this.

hannahker (approved):

Issue #74, opened 2024-05-13
Throw explicit errors if base maps are not covering everywhere we want to plot


caldwellst:
So, in `src-static/update_iso3_sf.R`, we develop base maps stored in `input/adm0/{iso3}.geojson`. These are intended to be the base map for any maps produced in the system, such as for mapping of points done in `map_points.R` or map of the IPC classifications in admin areas.

However, for some of the basemaps, country territories have been removed to ensure the visuals are of a better size. For instance, the Galápagos Islands are not included for Ecuador, and we only have the contiguous United States. This is mostly for island and some coastal countries. 

In most cases, this should not present issues. However, we should:

- Develop a function that checks that for any spatial data, its bounding box fits within the bounding box of the adm0 base map. We might want to use buffers here, because slight misalignment of polygon boundaries should not flag an error, we can't be expected to perfectly align all boundaries.
- Implement this function in `map_points.R` and any other indicator mapping functions that don't use `map_points.R`

Then, if we ever generate errors, we will have to develop custom basemaps for these situations that can then be reused in the future. For instance, if there is a tropical storm in Hawaii, we could develop a USA-H basemap that is used based on the spatial coordinates of the data. In this way, we don't have to deal with all eventualities, just deal with them as they arise, but confidently have the checks in place to know that we are not excluding points from our map.

zackarno:
Hey @caldwellst  - I created a new branch where I migrated the function I had previously written to check overlap of 2 geom bbox's. Here it is currently:  [assert_bbox_overlap](https://github.com/OCHA-DAP/hdx-signals/blob/assert-geom-overlap/src/utils/assertions.R) 

I'm trying to figure out where you want the function, 1. saved , 2. implemented.

On slack you said:

> Set it as a validation/assetion with gg_adm0 I think.

But  I don't think `gg_adm0()` is a thing 😄 . The most logical place i see for it's implementation, would `map_points()` ? However, the way the adm0 loading is wrapped in a `geom_*()` function makes it a bit clunky/inefficient? I think you will know what I mean, but let me know if not clear

caldwellst:
Implement in `geom_adm0()` I meant. So if you try to call it to use the base plot, you just pass in the other spatial files that will be plotted as `...`, and those all get checked against the `sf_adm0`. Feel free to pose alternative, but I think this makes most sense because we don't need multiline load and then plot the basemap in each plotting function, so validating at point of use in `geom_adm0` makes most sense.

[...]

PR #113, opened 2024-05-23
Assert geom overlap

zackarno:
I've implemented the function/assertion as discussed (mostly) #74 and figured it would make sense to move discussion here for final details

I wasn't exactly sure what you meant by using ... for other spatial files:

> So if you try to call it to use the base plot, you just pass in the other spatial files that will be plotted as ..., and those all get checked against the sf_adm0

were you thinking i'd send `...` to a list and then purrr$map() iterate the assertion over them? For now, I've just used the `additional_geom` argument with `default = NULL` as a place holder that works, but happy to adjust to the `...` if you can clarify.

I was thinking it would make sense to go in `map_points()` because i thought that would be where the point files are added to the map as have done in testing


src/utils/assertions.R
```diff
@@ -0,0 +1,69 @@
...
+#'    y =alleghany_county,
+#'    dist = 1000
+#'    )
+#'  assert_within_distance(
+#'    pts_sampled_bbox,
+#'    y = alleghany_county,
+#'    dist = 1000
+#'    )
+assert_within_distance <- function(x, y, dist = 10000) {
+  m_within_dist <- sf$st_is_within_distance(
```
caldwellst, line 50: `sf$st_is_within_distance()` fails with polygons, because it merely asserts that boundaries are within a distance.

```r
library(dplyr)
library(sf)
library(units)

file_name <- system.file("shape/nc.shp", package="sf")
clay <- filter(nc_counties, NAME == "Clay")
macon <- filter(nc_counties, NAME == "Macon")

st_is_within_distance(clay, macon, dist = units::set_units(1, "meters"), sparse = FALSE)
```

That returns `TRUE`.

caldwellst (changes requested):
So, you can see the comments, code is looking nice but the function is wrong because `sf::st_within_distance()` is not really checking if one shape is contained within another. Just a few other comments in the code itself for now.

src/utils/assertions.R
```diff
@@ -0,0 +1,69 @@
+box::use(dplyr)
```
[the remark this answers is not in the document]
zackarno: good call

src/utils/assertions.R
zackarno: ah yes good catch. Forgot a bout polys. Checkout the quick fix here: [f49bdb2](https://github.com/OCHA-DAP/hdx-signals/pull/113/commits/f49bdb26ec2775d0bf63b10f0847a2656db33c8d). Seems to work well

src/images/maps/geom_adm0.R
```diff
@@ -10,13 +11,18 @@ box::use(../../utils/iso3_shift_longitude)
 #' an error if the returned data is `NULL` or a 0 row data frame.
 #'
 #' @param iso3 ISO3 code
+#' @param additional_geom sf class object with geometry (default = NULL)
 #'
 #' @returns Geom for the country boundaries
 #'
 #' @export
-geom_adm0 <- function(iso3) {
+geom_adm0 <- function(iso3, additional_geom = NULL) {
```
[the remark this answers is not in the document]
zackarno: so like this: [c5593cb](https://github.com/OCHA-DAP/hdx-signals/pull/113/commits/c5593cb0bcc1300babf5992b644cbc7a381f5744)

src/images/maps/geom_adm0.R
```diff
@@ -10,13 +11,18 @@ box::use(../../utils/iso3_shift_longitude)
...
   sf_adm0 <- get_iso3_sf$get_iso3_sf(iso3, "adm0")
 
+  if (!null(other_geom)) {
```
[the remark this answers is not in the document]
zackarno: thanks good catch - fixed

src/images/maps/geom_adm0.R
```diff
@@ -34,3 +38,75 @@ geom_adm0 <- function(iso3, additional_geom = NULL) {
...
+    sparse = FALSE
+  )
+
+  not_within <- any(!m_within_dist)
+
+  if (not_within) {
+    idx_not_within <- which(!m_within_dist)
+
+    stop(
+      "Error: elements in x not contained within ", dist, "m of y:\n",
```
caldwellst, line 107: Change error message to be about the base map explicitly.

src/images/maps/geom_adm0.R
```diff
@@ -1,22 +1,37 @@
...
 #'
 #' @export
-geom_adm0 <- function(iso3) {
+geom_adm0 <- function(iso3, ...) {
   sf_adm0 <- get_iso3_sf$get_iso3_sf(iso3, "adm0")
 
+
+  additional_geoms <- list(...)
+  if (length(additional_geoms) > 0) {
+    purrr$map(
```
caldwellst, line 29: `purrr$walk` because we don't want return values, this is for the side effect of assertion.

src/images/maps/geom_adm0.R
```diff
@@ -25,6 +40,75 @@ geom_adm0 <- function(iso3) {
...
+#'    )
+#'  assert_within_distance(
+#'    pts_sampled_bbox,
+#'    y = alleghany_county,
+#'    dist = 1000
+#'    )
+assert_within_distance <- function(x, y, dist = 10000) {
+
+  # cast to points so a polygon w/ vertices outside `y` will create error
+  x <- sf$st_cast(x, "POINT")
```
caldwellst, line 92: You get warning messages here about repeating attrition utes for sub-geometries.

src/images/maps/geom_adm0.R
```diff
@@ -25,6 +40,75 @@ geom_adm0 <- function(iso3) {
...
+#' @param x sf class with geometry
+#' @param y sf class with geometry
+#' @param dist distance in meters (default = 10000)
+#'
+#' @return TRUE if all elements of x fall within `dist` distance of y.
+#'     Return error if any elements fall outside of distance
+#' @export
+#'
+#' @examples
+#' library(sf)
```
caldwellst, line 57: These examples won't work because you have no actual box imports. I think we can keep them but just note testing will be a much more complex setup.

src/images/maps/geom_adm0.R
```diff
@@ -1,22 +1,37 @@
...
+
+  additional_geoms <- list(...)
+  if (length(additional_geoms) > 0) {
+    purrr$map(
+      .x = additional_geoms,
+      .f = \(x) assert_within_distance(x, sf_adm0)
+    )
+  }
+
   if (is.null(sf_adm0) || nrow(sf_adm0) == 0) {
```
caldwellst, line 35: You need to run your assertion after we check that `sf_adm0` isn't `NULL` or missing, because yours will fail in these cases.

src/images/maps/geom_adm0.R
zackarno: good call

src/images/maps/geom_adm0.R
zackarno: true

src/images/maps/geom_adm0.R
zackarno: yeah ok. I like have the examples there as you can easily run through them (even in roxygen format!) manually and see the functions in action.

But yeah when it comes using for actual testing/CI im not sure how it fits into this repo design/box etc. I had just taken use of `{box}` out of the roxygen examples based on this comment: 
https://github.com/OCHA-DAP/hdx-signals/pull/75#discussion_r1598621531

but for the manual testing benefit box vs no box doesnt matter much

zackarno:
> Alright, just some final changes. I think my question is, will this actually be faster than just doing:
> 
> ```r
> st_covers(
>   x = st_buffer(sf_adm0, dist = set_units(10000, meters)),
>   y = sf
> )
> ```
> 
> Especially since we are going to be comparing polygons, casting complex subnational administration boundaries to points sounds like an expensive operation. What do you think?

- `st_covers()` could be nice - good find. Will look into it.
-  Yes it does casting to POINTS to does "sound" expensive, but based on  testing so far its been relatively instant on my computer.  I guess it's not such a hard operation as the points are all there w/ out any transformation already (just vertices).

The main benefit of the original is that, a. I've tested it visually and can see it's doing what we want. Nonetheless will try to see how `st_covers()` compares and make sure its straightforward as well.

zackarno:
annoyingly i just realized the last 2 comparisons evaluate differently! I guess i have to look into this

zackarno:
The issue was that `st_covers()` was doing the inverse of what I wanted in the way i had constructed the assertion. I've switched to `st_covers_by()` in the last 2 commits and we should be good to go now. I Some notes:

- using st_covers_by() is faster than casting the basemap to POINT and running `st_within_distance()`
- definitely many edge cases where the methods are not the same due to rounding by `st_buffer()`, but i think it definitely not important for us.
- the one case that won't work is where both the basemap and additional geom are POINTS. But I don't think we need to worry about this as the basemap will always be a poly right? If necessary we can do something clever to make it work, but will add a bit of complexity as it's not inherent to `st_covers_by()`

src/images/maps/geom_adm0.R
zackarno: fixed  [6ebb2b7](https://github.com/OCHA-DAP/hdx-signals/pull/113/commits/6ebb2b7f07bcfc994dcf49c84457c407b097545b)

src/images/maps/geom_adm0.R
zackarno: fixed by removal of st_cast

src/images/maps/geom_adm0.R
zackarno: implemented

src/images/maps/geom_adm0.R
zackarno: implemented

caldwellst:
Excellent work man! Will review now. I think you're right about it not mattering, and what we will do is ensure everything is a polygon in your PR to update the base maps. There is ONE that is points data I believe.

src/images/maps/geom_adm0.R
```diff
@@ -26,5 +32,87 @@ geom_adm0 <- function(iso3) {
...
+      .f = \(x) assert_covered_by(x, sf_adm0)
+    )
+  }
+
   gg$geom_sf(data = sf_adm0)
 }
+
+#' Assert that `x` is contained within `y`.
+#'
+#' Applies a buffer of `dist` meters to to polygon `x` and check if all
```
caldwellst, line 48: Buffering polygon `y`.

caldwellst (approved):
So good, thanks man, all looks great to me! The only comment is one documentation change, just a single letter difference. Then merge!

src/images/maps/geom_adm0.R
zackarno: sahweet! fixed

Issue #114, opened 2024-05-23
intentionally downloading and merging lowest level admin to create admin 0?


zackarno:
You can see that the `download_adm0_sf()` function does not actually download the admin 0 when `layer= NULL` (default). Rather it downloads all the admin files available and just selects the first one. So for example if you run 

```
download_adm0_sf(iso="BGD,file ="adm0")
```

it actually returns admin 4 level shapefile with > 5000 rows

https://github.com/OCHA-DAP/hdx-signals/blob/ab74f4a9181b71121fd01f3a5305f7dc9eb0188e/src-static/update_iso3_sf.R#L151-L187

You can see  later in the implementation of update_adm_sf .. layer is not specified either so it's downloading all the admins

https://github.com/OCHA-DAP/hdx-signals/blob/ab74f4a9181b71121fd01f3a5305f7dc9eb0188e/src-static/update_iso3_sf.R#L41-L43

The `update_adm_sf()` func does sort of convert them back to adm0 with this step:

https://github.com/OCHA-DAP/hdx-signals/blob/ab74f4a9181b71121fd01f3a5305f7dc9eb0188e/src-static/update_iso3_sf.R#L45-L48

but it seems like it would be cleaner just to download the admin 0 file as the `download_adm0_sf()` function name indicates. It might be possible that this unioning step is also resulting in some of the weirder looking shapes? 

Is the issue that the admin names are not standard enough? they look pretty standard just testing a few with `fieldmaps` so that you could string match the layer name on "adm\\d$" or supply it dynamically with `glue("{tolower(iso3)}_adm0")`. IF not standard enough you could just get the adm0 like this by grabbing the layer with the least features like this:

```
sf$st_layers(fn)$name[sf$st_layers(fn)$features == min(sf$st_layers(fn)$features)]
```

caldwellst:
Nice. I like this a lot, please implement, and yes, best to grab that adm0 layer!

zackarno:
do you happen to know if all FieldMap layer names are standardized to `glue("{tolower(iso3)}_adm\\d$")`.  If thats the case it's easy to supply as `layer` to `st_read()` inside `download_shapefile()`. If there not standardizing -- kind of got to take apart the `download_shapefile()` function to use the approach here:

```
adm0_lyr_name <- sf$st_layers(fn)$name[sf$st_layers(fn)$features == min(sf$st_layers(fn)$features)]
```

[...]

Issue #111, opened 2024-05-23
tag admin boundaries with `data_source` column attribute


zackarno:
Think we should tag the admin boundary with a data_source column attribute indicating where it's from. This can then be used in plot caption/legend/annotation. Also will be handy for debugging issues boundaries (ahem NZ).

probably could make sense to incorporate into `download_adm0_sf()` ? 

related to:
https://github.com/OCHA-DAP/hdx-signals/issues/87#issue-2298321368

caldwellst:
Yep, good idea! Could also track if we do or do not use a cities geom in the plot to refer to natural earth.

zackarno:
implementing an approach that could make sense here:
Fixes #115

PR #115, opened 2024-05-23
init data_source attribution for adm0

zackarno:
Not actually ready to be merged, but I thought rather than write a bunch of words in issue #111 , easier to just to just write some starter code in a branch to help discuss the approach. Happy to completely adjust approach and/or delete branch as we see fit.

You'll see I just implemented the approach for the Fieldmaps admin 0s as well as the first exception `iso3=ABA` (just put in a placeholder attribution for ABA). If we think it makes sense i can manually enter the attributions for the rest of the exceptions.

Will need to adjust this code potentially as/if we address #114 as well, but it should work in current construction of updating admins

### DIFF
--- a/src-static/update_iso3_sf.R
+++ b/src-static/update_iso3_sf.R
@@ -59,7 +59,7 @@ update_adm0_sf <- function(iso3) {
 #' @returns sf class object contain admin 0 country boundary
 #'
 #' @export
-simplify_adm0 <- function(iso3){
+simplify_adm0 <- function(iso3) {
   sf_adm0 <- download_adm0_sf(iso3) |>
     filter_adm0_sf(iso3)
 
@@ -100,7 +100,7 @@ simplify_adm0 <- function(iso3){
 #'
 #' @returns Filtered `sf_adm0`
 filter_adm0_sf <- function(sf_adm0, iso3) {
-  if (iso3 == "CHL"){
+  if (iso3 == "CHL") {
     sf_adm0 <- st_crop_adj_bbox(sf_adm0, xmin = 33.73339)
   } else if (iso3 == "BMU") {
     sf_adm0 <- st_crop_adj_bbox(sf_adm0, ymax = -0.05)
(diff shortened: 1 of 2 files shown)


src-static/update_iso3_sf.R
```diff
@@ -43,9 +43,14 @@ update_adm0_sf <- function(iso3) {
...
-    sf$st_union(sf_adm0) |>
-      sf$st_as_sf() # so we can check number of rows
+    sf_adm0 |>
+      # keepig it tidy allows us to retain cols
+      # i might even suggest mutating iso3 and grouping
+      # that in so it's also included in the file
+      dplyr$group_by(data_source) |>
+      dplyr$summarise(do_union = TRUE)
   )
+  # i did some speed benchmarking this against sf$st_union() and they are equivalent in speed.
```
caldwellst, line 53: Really cool. I like that we can just stick to tidy summarisation!

src-static/update_iso3_sf.R
```diff
@@ -195,7 +201,8 @@ download_adm0_sf <- function(iso3) {
 download_fieldmaps_sf <- function(iso3) {
   iso3 <- tolower(iso3)
   download_shapefile(
-    url = glue$glue("https://data.fieldmaps.io/cod/originals/{iso3}.gpkg.zip")
+    url = glue$glue("https://data.fieldmaps.io/cod/originals/{iso3}.gpkg.zip"),
+    data_source_label = "FieldMaps, geoBoundaries, U.S. Department of State, U.S. Geological Survey"
```
caldwellst, line 205: We are going to have to find ways to hide these because they are way too long.

src/utils/download_shapefile.R
```diff
@@ -45,15 +50,22 @@ download_shapefile <- function(url, layer = NULL) {
...
       quiet = TRUE
     )
   }
+  if (!is.null(data_source_label)) {
+    ret <- ret |>
+      dplyr$mutate(
+        data_source = data_source_label
+      )
+  }
+  return(ret)
```
caldwellst, line 70: No need for `return()`.

src-static/update_iso3_sf.R
```diff
@@ -157,7 +162,8 @@ download_adm0_sf <- function(iso3) {
   } else if (iso3 == "AB9") {
     download_shapefile(
       url = "https://open.africa/dataset/56d1d233-0298-4b6a-8397-d0233a1509aa/resource/76c698c9-e282-4d00-9202-42bcd908535b/download/ssd_admbnda_abyei_imwg_nbs_20180401.zip", # nolint
-      layer = "ssd_admbnda_abyei_imwg_nbs_20180401"
+      layer = "ssd_admbnda_abyei_imwg_nbs_20180401",
+      data_source_label = "Open Africa"
```
caldwellst, line 166: Maybe just `data_source`

caldwellst (commented):
Looks good for now, will review more when ready! I think what we might do is have a way to succinctly reference data sources in an image, because current ones much too long, and have a dataset somewhere that includes that information. So maybe add a `data_source_abbr` that is a very succinct version, and if we can't shorten enough, we can revisit the issue. We can include the full `data_source` in a country file that has all the metadata for our countries.

zackarno:
@caldwellst - This should be ready for review now.

Although it is related, I do not think we should include the switch from UN Geo Data Hub -> GAUL/Naturalearth in the PR as it's actually really complicated. Therefore would be good to get what we have attributed and deal with that as a separate issue.

I went ahead and addressed #125 here because it was simple. Ideally the LAC countries would come from FieldMaps however, but to do this properly we would need to significantly restructure most of the functions in update_iso3_sf.R. The issue is that sf_world is read in as one object while iso3s from other sources are looped/mapped.

caldwellst (approved):
Looks great! Just a few things to check in the comments.

zackarno:
Thanks for the review: all addressed [4882cbe](https://github.com/OCHA-DAP/hdx-signals/pull/115/commits/4882cbe997203f2cc1fb68e8a80e295fa50bf43a)

Should i try to run the update w/ the env var set so that it actually updates on the cloud? Or do you want to? Anything to be aware of? Should we archive the old files/store this in a different place?

caldwellst:
Update them all! No need to archive, we will want to use this new solution and can fix any issues as they arise.

PR #118, opened 2024-05-27
Hotfix to IDMC because hardcoded indicator_id incorrect

caldwellst:
Hardcoded `indicator_id` meant that both disaster related displacement and conflict related displacement were being assigned to conflict related.


hannahker (approved):
This fix looks good to me! 

This is something that I think we should improve down the line, as I've also been trying to index through all indicators by their `indicator_id` and this IDMC case makes things a bit messy.

PR #119, opened 2024-05-27
Remove overwrite_content functionality

caldwellst:
Previously, `overwrite_content` was a confusing parameter that when passed, meant that you expected existing rows of data in the signals data frame, and you would use those rows to generate new campaign content and signals. That is, you could basically not recalculate when you alert, just adjust how the content looked.

Turns out, this was way too complex, especially since recalculating when we want to alert is simple and easy and local, whereas the expensive computations are already in the content generation side. I've thus removed all of this functionality, and instead we just expect that if a dataset exists in `output/indicator/signals.parquet`, that it must be triaged before any new runs can happen. And if we elect to delete during triage, instead of deleting the content columns, it removes all rows so the entire data needs to be recreated. Much simpler this way I think.

### DIFF
--- a/src/alerts/check_existing_signals.R
+++ b/src/alerts/check_existing_signals.R
@@ -18,22 +18,19 @@ box::use(../utils/gmas_test_run)
 #'
 #' @param indicator_id ID of the indicator
 #' @param first_run Whether or not this is the first run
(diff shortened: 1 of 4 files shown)


hannahker (approved):
This makes me realize I misunderstood what `overwrite_content` was actually doing 😆 So definitely in favour of this to simplify things!

PR #120, opened 2024-05-27
Slack integration

### LINKED ISSUE #64: Use Slack notifications to alert when data needs triaging
When an indicator run happens and signals are generated, we have to manually triage the campaign and approve it before the email is sent and the data is added to the final file. One way could be to have a Slack channel that gets pinged automatically via the Slack API any time an indicator creates campaign information needing triage. Could use the [`{slackr}`](https://github.com/mrkaye97/slackr) package.

hannahker:
This PR introduces new functionality to output daily Slack updates on whether any new signals were detected and on the status of the monitoring scripts running in GitHub Actions. See screenshot below for examples of what these messages will look like, or #hdx-signals-bot-testing. 

Closes #64. 

If there are signals: 

Screenshot 2024-05-29 at 12 27 45 PM

If there are no signals: 

Screenshot 2024-05-29 at 11 57 53 AM

### DIFF
--- a/src/alerts/check_signals_updates.R
+++ b/src/alerts/check_signals_updates.R
@@ -101,6 +101,31 @@ slack_build_alert <- function(indicator_id, df) {
   )
 }
 
+#' Returns a DataFrame of all workflow runs on GitHub actions
+#' for monitoring a given indicator
+#'
+#' @param indicator_id ID of the indicator
+#'
+#' @returns DataFrame with metadata for all runs
+query_github <- function(indicator_id) {
+  workflow_id <- paste0("monitor_", indicator_id, ".yaml")
+  base_logs_url <- "https://github.com/ocha-dap/hdx-signals/actions/runs/"
+  httr2$request(
+    "https://api.github.com/repos/ocha-dap/hdx-signals/actions/workflows"
+  ) |>
+    httr2$req_url_path_append(
+      workflow_id,
+      "runs"
+    ) |>
+    httr2$req_auth_bearer_token(
+      token = get_env("GH_TOKEN")
+    ) |>
+    httr2$req_perform() |>
+    httr2$resp_body_string() |>
+    jsonlite$fromJSON(flatten = TRUE) |>
+    as.data.frame()
+}
+
 #' Takes the response from a GitHub Actions run of a single indicator
 #' and outputs a status message to be posted to Slack
 #'
@@ -108,24 +133,8 @@ slack_build_alert <- function(indicator_id, df) {
 #'
 #' @returns String status message to be posted to Slack
 slack_build_workflow_status <- function(indicator_id) {
-  workflow_id <- paste0("monitor_", indicator_id, ".yaml")
-  base_logs_url <- "https://github.com/ocha-dap/hdx-signals/actions/runs/"
-
   df_runs <- tryCatch({
-    httr2$request(
-      "https://api.github.com/repos/ocha-dap/hdx-signals/actions/workflows"
-    ) |>
-      httr2$req_url_path_append(
-        workflow_id,
-        "runs"
-      ) |>
-      httr2$req_auth_bearer_token(
-        token = get_env("GH_TOKEN")
-      ) |>
-      httr2$req_perform() |>
-      httr2$resp_body_string() |>
-      jsonlite$fromJSON(flatten = TRUE) |>
-      as.data.frame()
+    query_github(indicator_id)
   },
   error = function(e) {
     logger$log_error(e$message)
--- a/src/alerts/check_signals_updates.R
+++ b/src/alerts/check_signals_updates.R
@@ -51,9 +51,13 @@ slack_post_message <- function(header_text, status_text, signals_text) {
       list(type = "divider")
     )
   )
-  json_body <- jsonlite$toJSON(msg, pretty = TRUE, auto_unbox = TRUE)
-  response <- POST(get_env("SLACK_URL"), body = json_body, encode = "json")
-  if (response$status != 200) {
+
+  # Create and perform the POST request using httr2
+  response <- httr2$request(get_env("SLACK_URL")) |>
+    httr2$req_body_json(msg) |>
+    httr2$req_perform()
+
+  if (response$status_code != 200) {
     logger$log_error("Error posting Slack message")
     stop()
   }
--- a/src/alerts/check_signals_updates.R
+++ b/src/alerts/check_signals_updates.R
@@ -102,13 +102,12 @@ slack_build_header <- function(n_signals) {
 
 #' Builds the signal alert text
 #'
-#' @param iso3 Country iso3 code
 #' @param indicator_id Indicator ID
-#' @param campaign_url URL of Mailchimp campaign to review
+#' @param df DataFrame with all the signals for the given indicator
 #'
 #' @returns String signal alert text
-slack_build_alert <- function(iso3, indicator_id, campaign_url) {
-  return(paste0(iso3, ": ", indicator_id, " <", campaign_url, " | See draft campaign>\n"))
+slack_build_alert <- function(indicator_id, df) {
+  paste0("*", indicator_id, "* - ", nrow(df), " countries impacted- <", df$campaign_url_archive[1], " | See draft campaign>\n")
 }
 
 #' Takes the response from a GitHub Actions run of a single indicator
@@ -206,16 +205,10 @@ for (ind in indicators) {
   )
   df <- cs$read_az_file(fn_signals)
   if (nrow(df) > 0) {
-    for (i in seq_len(nrow(df))) {
-      row <- df[i, ]
-      alert <- slack_build_alert(
-        row["iso3"],
-        row["indicator_name"],
-        row["campaign_url_email"]
-      )
-      signals <- paste0(signals, alert)
-    }
-    n_signals <- n_signals + nrow(df)
+    logger$log_info(paste0("Found signal for ", ind))
+    alert <- slack_build_alert(ind, df)
+    signals <- paste0(signals, alert)
+    n_signals <- n_signals + 1
   }
 }
 logger$log_info(paste0("Found ", n_signals, " signals"))
--- a/src/alerts/check_signals_updates.R
+++ b/src/alerts/check_signals_updates.R
@@ -1,20 +1,21 @@
-library(httr)
-library(jsonlite)
+box::use(jsonlite)
+box::use(httr)
 box::use(src/utils/get_env[get_env])
 box::use(logger[log_error])
 box::use(src/utils/hs_logger)
-box::use(src/alerts/triage_signals[get_signals_df])
+box::use(cs = src/utils/cloud_storage)
 
 hs_logger$configure_logger()
 
 org_name <- "ocha-dap"
 repo_name <- "hdx-signals"
-base_logs_url <- "https://github.com/OCHA-DAP/hdx-signals/actions/runs/"
 indicators_path <- "src/indicators"
 indicators <- list.files(indicators_path)[file.info(file.path(indicators_path, list.files(indicators_path)))$isdir]
 test <- Sys.getenv("TEST", unset = "FALSE")
 
-process_run_status <- function(response, ind, base_logs_url) {
+process_run_status <- function(response, ind) {
+  base_logs_url <- paste0("https://github.com/", org_name, "/", repo_name, "/actions/runs/")
+
   if (status_code(response) == 200) {
     content <- content(response, as = "text", encoding = "UTF-8")
     json_data <- fromJSON(content, flatten = TRUE)
@@ -85,7 +86,7 @@ post_slack_message <- function(status_text) {
       list(type = "divider")
     )
   )
-  json_body <- toJSON(msg, pretty = TRUE, auto_unbox = TRUE)
+  json_body <- jsonlite$toJSON(msg, pretty = TRUE, auto_unbox = TRUE)
   response <- POST(get_env("SLACK_URL"), body = json_body, encode = "json")
   if (response$status != 200) {
     log_error("Error posting Slack message")
@@ -95,21 +96,30 @@ post_slack_message <- function(status_text) {
 
 full_status <- ""
 for (ind in indicators) {
-  print(ind)
   gh_response <- gh_status(ind, org_name, repo_name)
-  workflow_status <- process_run_status(gh_response, ind, base_logs_url)
+  workflow_status <- process_run_status(gh_response, ind)
   full_status <- paste0(full_status, workflow_status)
 }
 
 post_slack_message(full_status)
 
-for (ind in indicators) {
-    fn_signals <- paste0(
-        "output/",
-        ind,
-        if (test) "/test" else "",
-        "/signals.parquet"
-    )
+n_signals <- 0
 
-    df <- get_signals_df(fn_signals)
+# TODO: Hard coding this because the IDMC names don't totally match the indicator_ids
+indicators_azure <- c("acled_conflict", "idmc_displacement_conflict", "idmc_displacement_disaster", "ipc_food_insecurity", "jrc_agricultural_hotspots")
+for (ind in indicators_azure) {
+  print(ind)
+  fn_signals <- paste0(
+    "output/",
+    ind,
+    if (test) "/test" else "",
+    "/signals.parquet"
+  )
+
+  print(fn_signals)
+
+  df <- cs$read_az_file(fn_signals)
+  print(nrow(df))
+  n_signals <- n_signals + nrow(df)
 }
+print(n_signals)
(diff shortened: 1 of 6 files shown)


src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,125 @@
...
+  gh_response <- gh_status(ind, org_name, repo_name)
+  workflow_status <- process_run_status(gh_response, ind)
+  full_status <- paste0(full_status, workflow_status)
+}
+
+post_slack_message(full_status)
+
+n_signals <- 0
+
+# TODO: Hard coding this because the IDMC names don't totally match the indicator_ids
```
caldwellst, line 108: The IDMC names don't match because in one of the updates merged to main `indicator_id` was hardcoded to `idmc_displacement_conflict` and then passed to both runs. This has been fixed in the merge now.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+#'
+#' NOTE: More data will be returned as each workflow is run more times. Perhaps won't scale well,
+#' depending on how many runs a workflow will have
+#'
+#' @param indicator_id ID of the indicator
+#'
+#' @returns JSON API response
+gh_status <- function(indicator_id) {
+  # TODO Temp cleaning before workflows are renamed to match indicator ids
+  ind <- sub("^[^_]*_", "", indicator_id)
```
caldwellst, line 24: `sub()` and the base family of regex functions are perfectly fine to work with. Just note we have `{stringr}` in our current imports, which aims to be a much cleaner family of functions for this purpose. Basically, it has clearer function naming, like `str_detect()`, `str_extract()`, `str_remove()`, and a consistent API of the vector argument passed first as `x`, `pattern` as the second, etc. I literally constantly try to remember which of `sub()`, `gsub()`, `grep()`, etc. do what, and always have to refer to the help docs for that + where I place the arguments in the calls!

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
+box::use(jsonlite[fromJSON, toJSON])
+box::use(httr[GET, POST, add_headers, status_code, content])
```
caldwellst, line 2: A bit of a pain, but I think I am going to propose that we don't import objects like above but instead import the module and the reference exported objects using `module$object` notation. If the module is used frequently or has a long name, you could alias it like we do with `../utils/cloud_storage`.

caldwellst: As well, recommend to shift away from `{httr}` which has been superseded by `{httr2}`. Will add some rough attempt to convert the functions (without testing) in your code below!

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+#'
+#' @param indicator_id ID of the indicator
+#'
+#' @returns JSON API response
+gh_status <- function(indicator_id) {
+  # TODO Temp cleaning before workflows are renamed to match indicator ids
+  ind <- sub("^[^_]*_", "", indicator_id)
+  workflow_id <- paste0("monitor_", ind, ".yaml")
+  token <- get_env("GH_TOKEN")
+  url <- paste0("https://api.github.com/repos/ocha-dap/hdx-signals/actions/workflows/", workflow_id, "/runs")
```
caldwellst, line 27: So, it would be something like this in `{httr2}`:

```r
httr2$request(
   "https://api.github.com/repos/ocha-dap/hdx-signals/actions/workflows"
  ) |>
  httr2$url_path_append(
    workflow_id,
    "runs"
  ) |>
  httr2$req_auth_bearer_token(
    token = get_env("GH_TOKEN")
  ) |>
  httr2$req_perform() # can use httr2$req_method("GET") if not defaulting to GET, although I think it should
```

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+          )
+        )
+      ),
+      list(type = "divider")
+    )
+  )
+  json_body <- toJSON(msg, pretty = TRUE, auto_unbox = TRUE)
+  response <- POST(get_env("SLACK_URL"), body = json_body, encode = "json")
+  if (response$status != 200) {
+    log_error("Error posting Slack message")
```
caldwellst, line 75: With `{httr2}`, you can capture errors by adding `httr2$req_error()` in the pipe above, and wrapping there.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+            type = "mrkdwn",
+            text = status_text
+          )
+        )
+      ),
+      list(type = "divider")
+    )
+  )
+  json_body <- toJSON(msg, pretty = TRUE, auto_unbox = TRUE)
+  response <- POST(get_env("SLACK_URL"), body = json_body, encode = "json")
```
caldwellst, line 73: Pedantic, but should we do `HS_SLACK_URL`?

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+#' Builds the header text, depending on how many signals are reported
+#'
+#' @param n_signals Number of signals reported
+#'
+#' @returns String header text
+slack_build_header <- function(n_signals) {
+  if (n_signals == 0) {
+    title <- paste0(Sys.Date(), ": No signals identified")
+  } else {
+    title <- paste0(":rotating_light: <!channel> ", Sys.Date(), ": ", n_signals, " alerts identified")
```
caldwellst, line 89: So cool!

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+  }
+  return(title)
```
caldwellst, line 91: In the tidyverse style, no need to use `return()` if the value is the end of the function. In fact, you could literally just call `paste0()` without assigning it in both of your calls, and those would be the return values. Only use `return()` if exiting a function early, like this completely stupid example.

```r
a <- function(x) {
  if (all(is.na(x)) {
    return(1)
  }

  y <- x + 1
  x * y
}
```

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+
+#' Builds the signal alert text
+#'
+#' @param iso3 Country iso3 code
+#' @param indicator_id Indicator ID
+#' @param campaign_url URL of Mailchimp campaign to review
+#'
+#' @returns String signal alert text
+slack_build_alert <- function(iso3, indicator_id, campaign_url) {
+  return(paste0(iso3, ": ", indicator_id, " <", campaign_url, " | See draft campaign>\n"))
```
caldwellst, line 102: Same on `return()`

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+
+#' Takes the response from a GitHub Actions run of a single indicator
+#' and outputs a status message to be posted to Slack
+#'
+#' @param response JSON payload returned from `gh_status`
+#' @param indicator_id ID of the indicator
+#'
+#' @returns String status message to be posted to Slack
+slack_build_workflow_status <- function(response, indicator_id) {
+  base_logs_url <- paste0("https://github.com/ocha-dap/hdx-signals/actions/runs/")
```
caldwellst, line 113: No pasting done, can just assign the string on its own.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+for (ind in indicators) {
+  gh_response <- gh_status(ind)
+  workflow_status <- slack_build_workflow_status(gh_response, ind)
+  full_status <- paste0(full_status, workflow_status)
+}
+log_info("Successfully checked")
+
+log_info("Checking for signals across all indicators...")
+for (ind in indicators_azure) {
+  test <- Sys.getenv("TEST", unset = TRUE)
```
caldwellst, line 182: Can set outside the loop.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+  fn_signals <- paste0(
```
caldwellst, line 183: Maybe we should functionalise this since we are doing this simple paste a lot? Put into `utils`?

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,207 @@
...
+    "output/",
+    ind,
+    if (test) "/test" else "",
+    "/signals.parquet"
+  )
+  df <- cs$read_az_file(fn_signals)
+  if (nrow(df) > 0) {
```
caldwellst, line 190: In our runs, all rows of the campaign data frame will go into the same campaign, so a single URL. So maybe just need to get the # of campaigns and iso3 codes for the campaigns to report in the Slack message?

src/alerts/check_signals_updates.R
caldwellst: Also, depends on your style, but there are loads of options for string joining in R and this is one instance where you might prefer the `{glue}` style, which is basically like Python's f-strings.

```r
glue$glue(
  "{iso3}: {indicator_id} <{campaign_url} | See draft campaign>\n"
)
```

src/alerts/check_signals_updates.R
hannahker: Yes that looks so much better than regex! I also struggle big time. I'll leave for now since I'll remove this anyways once I rename all workflows to match the indicator IDs.

src/alerts/check_signals_updates.R
hannahker: Sounds good to me

src/alerts/check_signals_updates.R
hannahker: Ooh yeah this is nicer! I played around with it, but seems to have some complexity with the escape characters, like `\n` -- will defs keep in mind for the future

src/alerts/check_signals_updates.R
hannahker: Yup good call

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,193 @@
...
+indicators_files <- list.files(indicators_path)
+indicators_files <- indicators_files[stringr$str_starts(indicators_files, "monitor_")]
+indicators <- stringr$str_extract(indicators_files, "(?<=monitor_)[^\\.]+(?=\\.yaml)")
+
+full_status <- ""
+n_signals <- 0
+signals <- ""
+test <- Sys.getenv("TEST", unset = TRUE)
+
+for (ind in indicators) {
```
caldwellst, line 173: I have some questions about how we generally design the alert, and once resolved, will have some comments on how we approach building the final loop. I think could be nice to have a simple `purrr::map()` along the indicators, but that's a bit pedantic, so let's see.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,193 @@
...
+# Get the indicator ids of all workflows
+indicators_path <- ".github/workflows"
+indicators_files <- list.files(indicators_path)
+indicators_files <- indicators_files[stringr$str_starts(indicators_files, "monitor_")]
+indicators <- stringr$str_extract(indicators_files, "(?<=monitor_)[^\\.]+(?=\\.yaml)")
+
+full_status <- ""
+n_signals <- 0
+signals <- ""
+test <- Sys.getenv("TEST", unset = TRUE)
```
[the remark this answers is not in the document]
caldwellst: Thinking on how this testing should be reflected. Currently, the alerts in the Slack channel do not say test. Maybe have them say  test somewhere. Oh, no! Maybe, if `HS_TEST` is `TRUE`, then messages are sent to `#hdx-signals-bot-testing` and otherwise `#hdx-signals-bot`. That way we can still test these messaging functions without disrupting a channel that should only be notified when critical things are necessary to be responded to.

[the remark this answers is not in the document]
caldwellst: The other issue is if `HS_TEST` is `TRUE`, it compares against test data being available, which is always available now as we wait for contributors to comment. Not necessarily a problem, but we can't really use `HS_TEST` to test how Slack notifications work against what's actually going on in the repo.

caldwellst (changes requested):
So good, and amazing to have it in Slack! My main question focuses on how we actually want messages to appear in Slack.

I see why we duplicate entries for when there are signals in an indicator. It might be slightly cleaner to basically only have 1 row for each indicator_id with the following order:

- green circle: # of signals found
- green circle: successful update with no signals
- grey bar: No scheduled update
- red circle: Failure

I could see issues coming up if signals are sitting there for a while, so maybe the way you did it makes sense, and likely isn't that critical as we won't have this happen to much. Let me know what you think! Honestly think I'm overtired at this point so I might be rambling, so feel free to ignore.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,194 @@
...
+#' Builds the header text, depending on how many signals are reported
+#'
+#' @param n_signals Number of signals reported
+#'
+#' @returns String header text
+slack_build_header <- function(n_signals) {
+  if (n_signals == 0) {
+    paste0(Sys.Date(), ": No signals identified")
+  } else {
+    paste0(":rotating_light: <!channel> ", Sys.Date(), ": ", n_signals, " alerts identified")
```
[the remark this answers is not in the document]
hannahker: Fixed!

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,193 @@
...
+    # If no scheduled runs happened off of main today
+  } else if (nrow(df_sel) == 0) {
+    paste0(":heavy_minus_sign: ", indicator_id, ": No scheduled update \n")
+  } else {
+    paste0(":red_circle: ", indicator_id, ": More than one scheduled run today \n")
+  }
+}
+
+# Get the indicator ids of all workflows
+indicators_path <- ".github/workflows"
```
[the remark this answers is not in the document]
hannahker: Nice! Yeah, this is super efficient. I still need to get the hang of reading piped operations like this, but I like it

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,193 @@
...
+
+for (ind in indicators) {
+  logger$log_info(paste0("Checking GitHub Actions status for ", ind, "..."))
+  workflow_status <- slack_build_workflow_status(ind)
+  full_status <- paste0(full_status, workflow_status)
+  logger$log_info("Successfully checked")
+
+  logger$log_info(paste0("Checking for signals: ", ind, "..."))
+  fn_signals <- cs$signals_path(ind, test)
+  df <- cs$read_az_file(fn_signals)
```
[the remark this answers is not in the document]
hannahker: That's good to note. My assumption was that the `signals.parquet` files would always be present, which I think we should stick with.

src/alerts/check_signals_updates.R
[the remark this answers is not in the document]
hannahker: I like the idea of the two channels! I can implement that logic here.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,194 @@
...
+  )
+
+  # Create and perform the POST request using httr2
+  response <- httr2$request(get_env("HS_SLACK_URL")) |>
+    httr2$req_body_json(msg) |>
+    httr2$req_perform()
+
+  if (response$status_code != 200) {
+    logger$log_error("Error posting Slack message")
+    stop()
```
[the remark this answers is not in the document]
hannahker: Great point! I didn't notice that `stop()` was actually returning a message too. To keep things simple, I just moved the error message to `stop()` and removed the `logger$log_errors()` line. I spent some time Googling how these two could be combined, but didn't find an obvious solution and think it's probably not worth the time at this stage to perfect the formatting of this message.

src/alerts/check_signals_updates.R
```diff
@@ -0,0 +1,193 @@
...
+#' and outputs a status message to be posted to Slack
+#'
+#' @param indicator_id ID of the indicator
+#'
+#' @returns String status message to be posted to Slack
+slack_build_workflow_status <- function(indicator_id) {
+  workflow_id <- paste0("monitor_", indicator_id, ".yaml")
+  base_logs_url <- "https://github.com/ocha-dap/hdx-signals/actions/runs/"
+
+  df_runs <- tryCatch({
```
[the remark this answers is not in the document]
hannahker: I'm not totally following how this would be implemented with the `tryCatch()` flow. I've funtionalized the GH API call, which I think simplifies things. What do you think of the readability now?

hannahker:
> I see why we duplicate entries for when there are signals in an indicator. It might be slightly cleaner to basically only have 1 row for each indicator_id with the following order

Personally, I like having the alerts separate so that the messaging when there is a signal to triage looks notably different than the normal, non-signal updates. I think this will be more attention-grabbing for us when monitoring the channel. Curious to hear any other thoughts on how the layout could be improved!

caldwellst (approved):
I think looks good! We can merge now and if we do any design changes, can come later, but I like it as it is for now, and see your point about the alerts standing out.

PR #122, opened 2024-05-27
Hotfix signals return

caldwellst:
Fixes to some issues found in the code.

1. `log_info()` in `generate_signals()` was the final statement, so the logging was the return value, instead of the dataset itself. I didn't think I could find anywhere else where that error existed.
2. I think `test` being unset should always be `TRUE`. Runs should only be done intentionally. Are we sure we don't want `test` to be internal to the function, so we just use the env variable and remove the duplicated calls in each `update_{indicator}.R` script?
3. I have changed it so we only throw errors on warnings when not interactive. This should allow for a lot more flexibility when interactively developing. What do you think?
4. Some other small changes I notice when doing some runs.

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,2 +1,5 @@
 source("renv/activate.R")
-options(warn = 2)
+
+if (!interactive()) {
+  options(warn = 2)
+}
--- a/src/alerts/generate_signals.R
+++ b/src/alerts/generate_signals.R
@@ -179,7 +179,10 @@ generate_signals <- function(
     df_campaigns,
     fn_signals
   )
+
   log_info(paste0("Monitoring completed. ", nrow(df_campaigns), " signals generated for ", indicator_id))
+
+  df_campaigns
 }
 
 #' Validate campaigns data
--- a/src/indicators/idmc_displacement/update_displacement.R
+++ b/src/indicators/idmc_displacement/update_displacement.R
@@ -13,8 +13,9 @@ box::use(./utils/map_displacement)
 box::use(../../alerts/generate_signals[generate_signals])
 box::use(../../utils/hs_logger)
 
-test <- as.logical(Sys.getenv("HS_TEST", unset = FALSE))
+test <- as.logical(Sys.getenv("HS_TEST", unset = TRUE))
 test_filter <- if (test) c("AFG", "SSD") else NULL
+indicator_id <- "idmc_displacement_conflict"
 
 hs_logger$configure_logger()
 hs_logger$monitoring_log_setup(indicator_id)
@@ -26,7 +27,7 @@ df_wrangled <- wrangle_displacement$wrangle(df_raw)
 df_conflict <- generate_signals(
   df_wrangled = dplyr$filter(df_wrangled, displacement_type == "Conflict"),
   df_raw = dplyr$filter(df_raw, displacement_type == "Conflict"),
-  indicator_id = "idmc_displacement_conflict",
+  indicator_id = indicator_id,
   alert_fn = alert_displacement$alert,
   plot_fn = plot_displacement$plot,
   info_fn = info_displacement$info,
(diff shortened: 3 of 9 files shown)


src/alerts/generate_signals.R
```diff
@@ -179,7 +179,10 @@ generate_signals <- function(
     df_campaigns,
     fn_signals
   )
+
   log_info(paste0("Monitoring completed. ", nrow(df_campaigns), " signals generated for ", indicator_id))
+
+  df_campaigns
```
hannahker, line 185: Ahh interesting! I didn't know that the final line in a function was what was returned. I actually removed this return as it was adding a bit more clutter to the logs. Is there anywhere that this function needs to return anything, technically speaking?

.Rprofile
```diff
@@ -1,2 +1,5 @@
 source("renv/activate.R")
-options(warn = 2)
+
+if (!interactive()) {
```
hannahker, line 3: Nice 🌟

src/utils/get_env.R
```diff
@@ -14,10 +14,8 @@ get_env <- function(env_var_name, output = TRUE) {
   if (!nzchar(val)) {
     error_message <- sprintf("Environment variable '%s' is empty or not set.", env_var_name)
     log_error(error_message)
-    stop()
-  } else {
-    if (output) {
-      return(val)
-    }
+    stop(call. = FALSE)
```
hannahker, line 17: What is the `call = FALSE` doing here?

hannahker (approved):
Nice! Good finds! To this point...

> I think test being unset should always be TRUE. Runs should only be done intentionally. Are we sure we don't want test to be internal to the function, so we just use the env variable and remove the duplicated calls in each update_{indicator}.R script?

Yes, I do like cleaning things up to just reference the env variable from the function, instead of passing it through. I can work on that one!

src/alerts/generate_signals.R
caldwellst: Technically it doesn't, but it's nice when developing to be able to output the data frame and explore it. I think if it's cluttering the logs, the answer in R is always to assign the variable to something. So in the `update_{indicator}.R` scripts, we can make sure `generate_signals()` is assigned to an object, e.g. `df_blah <- generate_signals()`.

src/utils/get_env.R
hannahker: Love it! Thanks for clarifying

src/indicators/idmc_displacement/update_displacement.R
```diff
@@ -26,7 +28,7 @@ df_wrangled <- wrangle_displacement$wrangle(df_raw)
 df_conflict <- generate_signals(
   df_wrangled = dplyr$filter(df_wrangled, displacement_type == "Conflict"),
   df_raw = dplyr$filter(df_raw, displacement_type == "Conflict"),
-  indicator_id = "idmc_displacement_conflict",
```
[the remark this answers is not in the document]
caldwellst: Yes, that's perfectly fine! Want to create a separate PR for it? There is slight duplication in reading from the "API" (it's not a real API, it's just a URL they gave us that has data behind, so we sadly can't query it to my knowledge). But I think that's fine!

src/alerts/generate_signals.R
caldwellst: @hannahker Have pushed it so the `generate_signals()` calls are saved to a var. Let's see if that clears up the logs? Is there a way I can test before merging?

caldwellst:
Alright @hannahker, final two questions for you on the unresolved comments then I'll merge!

src/alerts/generate_signals.R
hannahker: Yup you can run a workflow off of a branch! 
Screenshot 2024-05-28 at 12 29 46 PM

src/indicators/idmc_displacement/update_displacement.R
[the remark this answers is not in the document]
hannahker: Great, I'll pick this up in another PR

PR #126, opened 2024-05-28
Use new logo without the pattern

caldwellst:
Simple shift to use new logo stored on Mailchimp.


zackarno:
LGTM - how can I preview this update myself again? Maybe you can walk us through this (off-GH)

zackarno (approved):

caldwellst:
You should be able to now run any of the `update_{indicator}.R` functions in test mode and it will open up a preview HTML. But probably a better way to do this in the long run to easily open up a preview! Maybe some quick batch scripts that run any of the actions in test mode?

PR #127, opened 2024-05-28
Setting plot and panel background to white

caldwellst:
Set plot and panel background to white. Transparent background ran into issues with dark mode. Not sure if I need to set both, but I believe `plot.background` is limited to plot extents, but `ggsave(..., bg = NULL)` is the default argument and when `bg` is `NULL`, it saves with the value specified as `plot.background`. So figured safer to set both.

### DIFF
--- a/src/images/plots/theme_signals.R
+++ b/src/images/plots/theme_signals.R
@@ -21,6 +21,8 @@ theme_signals <- function() {
       axis.text = gg$element_text(size = 11),
       plot.caption = gg$element_text(size = 11, hjust = 1),
       legend.text = gg$element_text(size = 9),
-      legend.title = gg$element_text(size = 11)
+      legend.title = gg$element_text(size = 11),
+      panel.background = gg$element_rect(fill = "white"),
+      plot.background = gg$element_rect(fill = "white")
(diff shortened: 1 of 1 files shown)


zackarno:
If I set toggle off `bg` specification below:

https://github.com/OCHA-DAP/hdx-signals/blob/47fd332810d3a941030b1ee5884ca35523a4bc52/test/manual/maps/map_test.R#L61-L68

and I run [test/manual/maps/implement_map_test.R](https://github.com/OCHA-DAP/hdx-signals/blob/plot-white-bg/test/manual/maps/implement_map_test.R) (on a subset), the pngs have a grey transparent bg. I dont think this makes sense. Can you try?

zackarno:
Actually sorry disregard the above. I restarted my R-session and the re-ran -- white background came through 😅 .

What is also weird is that the issue with the text-size  being smaller on my computer also seems to have gone away? what in the world?

zackarno (approved):

caldwellst:
Yeah we fixed that in a different branch basically by automatically running what you said to set the `{showtext}` DPI to 300!

Issue #94, opened 2024-05-17
Clarify `test` and `GMAS_TEST_RUN` variables


hannahker:
For simplicity's sake, can we consolidate a "testing" toggle into a single variable? Looking at the documentation [here](https://github.com/OCHA-DAP/hdx-signals/blob/main/src/alerts/generate_signals.R#L45-L52), it's a little confusing to me how these two interact. For example, if [this `test` flag](https://github.com/OCHA-DAP/hdx-signals/blob/52f083ec5eaa026ce1a8664cd812006be6604ebb/.github/workflows/monitor_conflict.yaml#L13-L19) is `TRUE` on a run, what exactly should happen?

caldwellst:
I'm leaning no, but happy to be convinced otherwise. I'll go through my reasoning and how I use them. Mainly it is because I think `GMAS_TEST_RUN` is not really a a test toggle as such, and could do with renaming, and I think having the 4 options they provide together are quite nice for development and testing. However, once described interested to hear how we could combine or simplify but still maintain the functionality they currently provide.

However, I'll let that all be up for discussion, and just describe how I think at least having the 4 toggles are very useful. Let's also use this to decide what to rename `GMAS_TEST_RUN` to.

So, I use `test` to automatically generate signals, without having to worry about if there is actually any new runs. It ensures we get signals to validate, by changing `filter_alerts()` to include latest signal from randomly sampled ISO3 values (or set ISO3 codes using `test_filter`) and bypasses all the checks and validation on existing data in the `signals.parquet` files. Otherwise, all the functions are designed to not run if there are no new signals, or existing signals, or any number of reasons. It's all about seeing the final content in the email and how it looks when sent, so it makes sure we have stuff that generates a campaign and allows us to preview it.

`GMAS_TEST_RUN` was originally just used to allow me to develop and test locally without any fear of messing things up or spending money. The only thing it really does is stop functions that save to the cloud, ping AI, and save to Mailchimp. It does this at the last minute, typically on calls to external functions that interact outside the repo.

This, I think?, makes it really nice in that you can basically do end to end test runs of the system, excluding the final saving functions (which we could unit test). However, I also really like it for local development.

So here's how the combination works, and how I find it useful:

- Both `TRUE`: No remote interactions to save data or send emails. Campaigns are guaranteed to be generated, even if data already exists or we otherwise wouldn't generate one.
- Both `FALSE`: Normal run.
[...]

PR #131, opened 2024-05-29
Add continuous testing

### LINKED ISSUE #110: Add continuous integration for monitoring scripts
Before merging to `main`, we should add automated checks to make sure that the `update_{indicator}.R` scripts all still run successfully without errors or warnings. Can just do the basic version of this, with `TEST = TRUE` and `GMAS_TEST_RUN = TRUE`.

hannahker:
Closes #110 and #94.

For far, working to add basic tests for all indicators. These tests should run without external API calls (eg. to IPC or Mailchimp APIs). However, we'll still need connections to the Prod Azure Blob to read in necessary static files for indicators (eg. geojsons for mapping). 

- [x] JRC Agricultural hotspots
- [ ] IPC Food insecurity - Runs with IPC API connection
- [ ] IDMC Conflict - Need sample data
- [ ] IDMC Displacement - Need sample data
- [ ] ACLED Conflict - Need sample data

This PR also includes changes to move globally defined datafames into functions, so that they won't be called unnecessarily by imports.

The tests for all indicators should be extended beyond the base cases added here. This can be done by running the `generate_signals_wrapper()` functions against different input raw datasets, which are generated manually and stored in `test/data/*.RDS`.


caldwellst:
> This setup is designed to make sure that 1) the generate_signals() scripts continue to run successfully for all indicators, and 2) the output of this function is consistent for all indicators. My understanding is that generate_signals() is a function that will pretty comprehensively run through a lot of the overall codebase, so continuously testing it for these various input indicators will be quite useful.

Makes sense!

> I've created new test_ scripts that closely mirror our existing update_ scripts. I prefer this so that we can keep the testing code isolated as much as possible. I think it also makes sense not to bundle this in with the existing test or GMAS_TEST_RUN variables since it'll be cleaner to keep those focused on enabling development (and not testing as well).

Makes sense to have separate scripts, to ensure that we can use the test data, and makes sense to have test scripts. A concern I have is say we make changes to an indicator run, such as adding in a second plot to the email. That function would need to be imported and passed to `generate_signals()`. Is there any way to ensure a match between `update_{indicator}.R` and `test_{indicator}.R`? Maybe we could enforce that if `update_{indicator}.R` is changed that the matching `test_{indicator}.R` file is changed? This is done often in R packages that won't let you commit changes to an `.Rmd` file if there is no change to the output `.md` file. It would be even better if we could enforce that just if there was changes to the `generate_signals()` call, but not sure that's possible?

Agree about keeping them separate! Maybe that helps us define what we call them? Because `test` is not about testing, but about doing a dry run to view the visual outputs? Could we call it `HS_DRY_RUN`? Does that make sense? And `HS_LOCAL` instead of `GMAS_TEST_RUN`?

> I'm imagining we run these scripts with test = FALSE and GMAS_TEST_RUN = TRUE -- so we mimic the "production" run of the code without connecting to external systems.

Agreed! However, I can see an issue with the idea to keep `GMAS_TEST_RUN` (will call it `HS_LOCAL` below). For the IPC food insecurity runs, maps are loaded in directly from the IPC API, because you can pull in geojson files for specific analyses. So if we wanted to truly make this local, would need to mock those files as well as the `raw_{indicator}.R` inputs.

And if `HS_LOCAL` is also used for development, in those cases, we WOULD want to call the IPC API to download the map, and run the `raw_{indicator}.R` file as normal against the API. So maybe `HS_LOCAL` sticks to ensuring the outputs are kept local. And mocking of the inputs happens for the `test_{indicator}.R` portions?

> I started looking into testthat, which looks like a good package to give us an over all unit testing framework. I haven't pursued it here further just because it'll take me longer to figure out and I want to prioritize getting some very basic tests in place first.

Yep it's really good! I think we can build on top of what you put in place to add it in later, as we can add more unit testing on other functions in `utils` and elsewhere. But don't think it's critical for what you're doing.

> The key piece of work here will be creating input dummy raw datasets for each indicator. I think these should be relatively small, but should cover as many edge cases as we can think of. I'm undecided on the best way to store this data. Should it be hard-coded dataframes in R? parquet files in the codebase? dummy files in Azure? I would lean away from external storage (like Azure) so that we're not potentially messing up our production data, and also because we don't really want to have potential cloud connection issues interfere with our testing.

Okay, I think good idea not to store up on Azure. My initial idea was to have scripts to generate test data by filtering real data to include all of the edge cases we want. This means we could regenerate easily if changes are done in the source data. It seems fairly common from what I've seen elsewhere to hardcode in scripts as you mentioned. However, some of the raw data frames are quite large in terms of columns, and might be a pain to maintain even if the number of rows are limited.

Actually, thinking about it, I think would be extremely difficult because to test the flagging methods for IDMC displacement and ACLED conflict, we need over 3 years of events data for a country, which will be a significant amount of data to hardcode. I think the filtration option would be the only possibility in this case I think?

.github/workflows/test_monitoring.yaml
```diff
@@ -0,0 +1,35 @@
...
+            run: |
+              shopt -s globstar
+              set -e
+              for x in test/automated/test_*.R; do Rscript "$x"; done
+            env:
+              HS_LOCAL: "TRUE"
+              DRY_RUN: "FALSE"
+              LOG_LEVEL: "ERROR"
+              DSCI_AZ_SAS_PROD: ${{ secrets.DSCI_AZ_SAS_PROD }}
+              IPC_API_KEY: ${{ secrets.IPC_API_KEY }}
```
hannahker, line 35: Will work to remove this

hannahker:
Closing in favour of #172

PR #132, opened 2024-05-30
Location metadata

caldwellst:
This is ready for review, but putting into draft as we still are awaiting language confirmation on a few things. Basically, need to put together this file that contains all the locations metadata including indicator coverage. However, at the same time realised that we need to move our naming of the countries we analyse from `country` to `location`. This is mainly a backend thing and not often pushed out to the user, but will impact how people sign up to Mailchimp.

Need to test how it performs, but will wait once I have final signup form language tomorrow. That is when we can better test performance.

## Coverage update

The primary functionality updated is that we use `update_coverage()` in all of the indicator updates to ensure that we keep an up to date list of places covered for that indicator in Azure. Then that is pulled into the metadata. Maybe that needs to be put into a scheduled update.


src-static/update_location_info.R
```diff
@@ -121,11 +130,14 @@ df_names <- df_ocha_names |>
...
   dplyr$transmute(
     iso3,
-    country,
-    unhcr_region = countrycode$countrycode(
-      sourcevar = iso3,
-      origin = "iso3c",
-      destination = "unhcr.region"
+    location,
+    location_note,
+    unhcr_region = suppressWarnings(
```
hannahker, line 135: What warnings are we suppressing here?

src/indicators/acled_conflict/update_conflict.R
```diff
@@ -1,4 +1,5 @@
 box::use(logger[log_info, log_debug])
+box::use(readxl)
```
hannahker, line 2: Was this missing before? Not seeing any use of this in the diff.

src/utils/ai_summarizer.R
```diff
@@ -9,31 +9,31 @@ box::use(../utils/hs_logger)
...
-ai_summarizer_without_country <- function(prompt, info, country) {
+ai_summarizer_without_location <- function(prompt, info, location) {
   ai_summary <- ai_summarizer(prompt = prompt, info = info)
-  if (stringr$str_detect(ai_summary, country)) {
+  if (stringr$str_detect(ai_summary, location)) {
     ai_summary <- ai_summarizer(
       prompt = paste0(
-        "Please rewrite this short text to exclude the country name ",
-        country,
+        "Please rewrite this short text to exclude the location name ",
```
hannahker, line 35: Curious... is it worth doing any testing to see if ChatGPT interprets "location" vs "country" differently? Does this change anything unexpected in the outputs?

hannahker:
Phew what a diff! We should definitely do some manual testing before merging to make sure there are no expected errors introduced (ie. remaining mismatches between `country` and `location`). Just left a couple clarifying questions. 

> Then that is pulled into the metadata.

What metadata do you mean?

src/indicators/acled_conflict/update_conflict.R
caldwellst: No, good catch! Was going to read an Excel file ACLED stores on their website which gives their coverage.

src-static/update_location_info.R
caldwellst: I'll do it properly was being a bit lazy. Basically there is an annoying thing where `ifelse()` is vectorized, so can be a nice way to merge two function calls that return vectors based on some logic. In this case, I was only calling `countrycode$countrycode()` if I knew I didn't have a custom region defined, because the custom region was set to deal with missing values in `countrycode()`! But, the way `ifelse` works, both of the potential outputs are evaluated in full, and it's only the return values that the boolean is applied to. So this generate warnings that there was no match to some of the ISO3 codes passed in, which we could safely ignore since we were dealing with those elsewhere. But to be cleaner I've instead parsed the code so we only pass in subsets of the vectors we know won't generate warnings either side.

```r
# don't use dplyr to avoid warnings for missing values in countrycode
uses_custom_reg <- !is.na(df_regions$region_custom)
df_regions$unhcr_region <- NA_character_
df_regions$unhcr_region[uses_custom_reg] <- df_regions$region_custom[uses_custom_reg]
df_regions$unhcr_region[!uses_custom_reg] <- countrycode$countrycode(
  sourcevar = df_regions$iso3[!uses_custom_reg],
  origin = "iso3c",
  destination = "unhcr.region"
)
```

caldwellst:
Alright, have tested this on all indicators and static update scripts, as well as listing, and we are successful! I think we can merge now because need to now change the whole design of the email and make final indicator changes ha...

src-static/update_acled_info.R
```diff
@@ -35,19 +35,19 @@ if (interactive()) {
...
       start_date = as.Date(
         x = paste(1, `Start Date`),
         format = "%d %B %Y"
       ),
-      acled_hdx_country = Country |>
+      acled_hdx_location = Country |>
         stringr$str_remove_all("[^[:alnum:]\\s-\\.]") |>
         stringr$str_remove_all("\\.$") |>
         tolower() |>
         stringr$str_replace_all("\\s|\\.", "-"),
```
zackarno, line 47: would this serve the same purpose more concisely:

```
stringr$str_replace_all(snakecase$to_snake_case(Country), "_", "-")
```

src-static/update_adm0_sf.R
```diff
@@ -42,7 +40,11 @@ suppressMessages(
 #'
 #' @export
 update_adm0_sf <- function(iso3) {
-  sf_simplified <- simplify_adm0(iso3)
+  sf_simplified <- download_adm0_sf(iso3) |>
```
zackarno, line 43: beauty

src-static/update_adm0_sf.R
```diff
@@ -157,63 +156,63 @@ st_crop_adj_bbox <- function(sf_obj, xmin = 0, xmax = 0, ymin = 0, ymax = 0) {
...
+download_adm0_sf <- function(iso3) {
   if (iso3 == "LAC") {
     # for LAC we get all 3 of El Salvador, Guatemala, and Honduras
     dplyr$filter(sf_world, iso3cd %in% c("SLV", "GTM", "HND")) |>
       dplyr$group_by(iso3cd) |>
       # pull together polygons by iso3cd
       dplyr$summarise(do_union = TRUE, .groups = "drop") |>
       # then merge all iso3cds into 1 multipolygon retaining boundaries
-      dplyr$summarise(do_union = FALSE)
+      dplyr$summarise(boundary_source = "UN Geo Hub", do_union = FALSE)
```
zackarno, line 178: cool didn' know adding the col would work as well

src-static/update_adm0_sf.R
```diff
@@ -157,63 +156,63 @@ st_crop_adj_bbox <- function(sf_obj, xmin = 0, xmax = 0, ymin = 0, ymax = 0) {
...
-      data_soure = "Who's On First"
+      boundary_source = "Who's On First"
     )
   } else if (iso3 == "IOT") {
-    download_shapefile(
+    download_shapefile$download_shapefile(
       url = "https://public.opendatasoft.com/api/explore/v2.1/catalog/datasets/world-administrative-boundaries/exports/geojson?lang=en&timezone=Europe%2FLondon",#nolint
-      data_source = "Opendatasoft"
+      boundary_source = "Opendatasoft"
     ) |> # nolint
```
zackarno, line 196: what will the lintr not like here?

src-static/update_acled_info.R
caldwellst: Possibly, but put in a separate PR for review ha!

src-static/update_adm0_sf.R
```diff
@@ -232,14 +231,14 @@ download_adm0_sf <- function(iso3, update_azure = TRUE) {
 #' @param iso3 ISO3 code
 download_fieldmaps_sf <- function(iso3, layer = NULL) {
   iso3 <- tolower(iso3)
-  download_shapefile(
+  download_shapefile$download_shapefile(
     url = glue$glue("https://data.fieldmaps.io/cod/originals/{iso3}.gpkg.zip"),
-    data_source = "FieldMaps",
+    boundary_source = "OCHA, Fieldmaps",
```
zackarno, line 236: Think capitalizations should be: FieldMaps based on full attribution [here](https://fieldmaps.io/data)

_Attribution: FieldMaps, geoBoundaries, U.S. Department of State, U.S. Geological Survey_

caldwellst: Aaah I see yes, strange that he refers to it as Fieldmaps elsewhere on the repo! Nice catch.

caldwellst: And will switch to be `Fieldmaps, OCHA`

src-static/update_adm0_sf.R
caldwellst: Dunno, wasn't added in this PR, I imagine maybe was kept when the `data_source` was added maybe if the line was previously one long line?

caldwellst: Have removed!

src-static/update_locations.R
```diff
@@ -114,56 +90,46 @@ region_match_df <- dplyr$tribble(
...
+if (any(is.na(dplyr$select(df_locations, -location_note)))) {
+  stop(
+    "Location date has missing values, fix before re-running.",
+    call. = FALSE
+  )
+} else {
+  fname <- "input/locations.parquet"
+  cs$update_az_file(
+    df = df_locations,
+    name = fname
```
zackarno, line 132: might be good to explicitly specify `blob` arg

src-static/update_locations_metadata.R
```diff
@@ -0,0 +1,94 @@
...
+
+fname <- "input/locations_metadata.parquet"
+cs$update_az_file(
+  df = df_locations_metadata,
+  name = fname
+)
+
+# TEMP: switch when system can read in parquet from Azure prod
+cs$update_az_file(
+  df = dplyr$select(df_locations_metadata, -lat, -lon),
```
zackarno, line 89: they dont need that lat/lon!?

src/images/maps/geom_adm0.R
```diff
@@ -95,7 +92,7 @@ geom_adm0 <- function(iso3, ...) {
 assert_covered_by <- function(x, y, dist = 10000) {
   y_buff <- sf$st_buffer(
     y,
-    dist = units::set_units(dist, metres)
+    dist = units$set_units(dist, metres)
```
zackarno, line 95: nice catch

src/utils/add_locations_metadata.R
```diff
@@ -0,0 +1,43 @@
+box::use(dplyr)
+box::use(stringr)
+
+box::use(cs = ./cloud_storage)
+
+#' Add location and region columns
```
zackarno, line 6: feel like it would be nice to add a "why" part to the explanation. And maybe that would include where the function is called

zackarno:
Nice looks great! 

haven't finished yet, but going to create a running list of some general stuff here so you will have some useful feedback tomorrow before I start

still getting this warning on when i run `update_displacement_*.R`
```
Warning message:
In st_centroid.sfc(x, of_largest_polygon = FALSE) :
  st_centroid does not give correct centroids for longitude/latitude data
```

src/utils/add_locations_metadata.R
zackarno: Also I think IRN `location` looks truncated: 

`Iran (Islamic Republic of)`

zackarno (approved):

zackarno:
Haven't fully gone through, but LGTM! and i think merging will expedite troubleshooting/fixes/iteration

src-static/update_locations_metadata.R
```diff
@@ -0,0 +1,94 @@
...
+  )
+
+#######################
+#### FINAL DATASET ####
+#######################
+
+df_locations_metadata <- purrr$reduce(
+  .x = list(df_locations, df_hrp, df_centroids, df_coverage),
+  .f = \(x, y) dplyr$left_join(x, y, by = "iso3")
+)
```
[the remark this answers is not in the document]
caldwellst: Ooh that's way nicer!

src-static/update_locations_metadata.R
caldwellst: We are not sharing `lat/lon` publicly, it's just country centroids, so only use for it is the internal mapping from Erika.

src-static/update_locations.R
caldwellst: I think on this we want to rely on the default being `prod` since we only don't want to read or write from `prod` in a very rare set of circumstances. In fact, don't even want to do that long term, ideally will only work to and from `prod`.

src/utils/add_locations_metadata.R
caldwellst: In the maps?

caldwellst:
> a quick "Find in files" (command + shift + F) search did reveal a few more places where country is still used over locations. Perhaps double check and decide if they need to change

Yep, good idea, removed a few more that came back when merging to main!

> still getting this warning on when i run `update_displacement_*.R`

Can you create an issue for this? Because there is no centroids call in there. We will hopefully catch anyway, but if you can show me will be easier to jump in and fix.

PR #134, opened 2024-05-31
Functionalize connections to Azure blob

hannahker:
Minor change so that we don't have to have the SAS tokens defined when not needed -- for example: https://github.com/OCHA-DAP/hdx-signals/actions/runs/9316443921/job/25644717361

The two functions are somewhat redundant, but I like the ease of calling just `blob_dev()`, for example, and not having to specify all the input parameters.


caldwellst (approved):
Ah smart approach!!

PR #151, opened 2024-06-10
Fix legends

### LINKED ISSUE #148: Check legend overlap in maps
In some instances, we get legend overlap in maps. 

Placement is defined in the assets in `update_iso3_map_settings.R`. We are now no longer using the dynamic sizing for maps (they are all 6x4 inches). It might be worth simply placing the legend at the bottom/top or left/right of a plot rather than placing within the plot for a few countries.

zackarno:
fixes #148.

as the the loading of map settings json from blob is essentially hardcoded into the plotting functions wondering the best way to develop test out the changes. 

I could overwrite the json, but worried of introducing a bigger issue. I think I can just temporarily edit the `map_theme.R` file to load in a local version of `iso3_map_settings.json`, but just checking if there would be a smoother way


zackarno:
I've been putting map outputs for visual inspection in the gdrive: `Shared drives/Predictive Analytics/HDX Signals/tmp`

It doesn't seem like fixing every concern with `breaks_pretty()` is simple ?breaks_pretty states : "You may get slightly more or fewer breaks that requested." in the arg description

caldwellst (approved):

caldwellst (approved):

PR #154, opened 2024-06-10
Add GitBook docs

### LINKED ISSUE #76: Build out documentation in GitBook
We want to build out documentation in GitBook. This should be relatively straightforward, but I expect there to be some hiccups. One thing I want to figure out alongside just having all documentation in GitBook is how we can embed some updatable figures in there for specific documentation on things like:

- Geographic coverage of datasets
- Expected frequency of emails based on historic data

I imagine simple enough to setup Rmds for them, and maybe have easy GitHub Actions to update them.

hannahker:
Closes #76

### DIFF
--- a/.gitbook.yaml
+++ b/.gitbook.yaml
@@ -1,0 +1,5 @@
+root: ./docs/
+
+​structure:
+  readme: README.md
+  summary: SUMMARY.md​
--- a/.gitbook.yaml
+++ b/.gitbook.yaml
@@ -1,5 +1,5 @@
 root: ./docs/
 
-​structure:
+structure:
   readme: README.md
-  summary: SUMMARY.md​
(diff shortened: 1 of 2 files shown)


.gitbook.yaml
```diff
@@ -0,0 +1,5 @@
+root: ./docs/
+
+​structure:
+  readme: README.md
+  summary: SUMMARY.md​
```
caldwellst, line 5: What's the strange parsing here indicating? Incorrect yaml?

caldwellst (approved):
Looks good!

.gitbook.yaml
hannahker: Speediest review!! My copy and paste from another file might have gone weird. Just pushed a fix.

PR #157, opened 2024-06-11
Technical docs

caldwellst:
Technical docs! Tried to at least get the top level directories started for your review. Good to finish these up and merge, then build in more detail as necessary so that we don't have the current docs sitting on `main`.


CONTRIBUTING.md
```diff
@@ -0,0 +1,98 @@
...
+If you want to make a bigger change, it's a good idea to first file an issue and make sure someone from the team agrees that it’s needed. 
+If you’ve found a bug, please file an issue that illustrates the bug with a minimal 
+[reprex](https://www.tidyverse.org/help/#reprex) (this will also help you write a unit test, if needed).
+See the Tidyverse guide on [how to create a great issue](https://code-review.tidyverse.org/issues/) for more advice.
+
+### Pull request process
+
+*   Pull the latest version of `main` and install all dependencies with `renv::restore()`.
+*   Create a Git branch for your pull request (PR). You could use `usethis::pr_init("brief-description-of-change")`.
+*   Make your changes, commit to git, and then create a PR by running `usethis::pr_push()`, and following the prompts in your browser.
```
hannahker, line 20: Would want to make it clear that `usethis` is optional and people can still use any other methods to create PRs.

CONTRIBUTING.md
```diff
@@ -0,0 +1,98 @@
...
+    Ensure that `CHANGES.md` is updated as described below.
+    The title of your PR should briefly describe the change.
+    The body of your PR should contain `Fixes #issue-number`.
+
+### Update `CHANGES.md`
```
hannahker, line 25: It's a bit hard to follow this section and understand the relationship between `CHANGES.md` and `NEWS.md`. So someone would edit `CHANGES`, which automatically populates the `NEWS` file?

I think having a very declarative, procedural intro sentence to sections could be helpful for people who will only skim-read the docs. Maybe something like: 

"Add a bullet point to the `CHANGES.md` file to describe the changes in your PR."

CONTRIBUTING.md
```diff
@@ -0,0 +1,98 @@
...
+*   `System`: changes in the `src` code that are not necessary to reflect publicly,
+which can include improvements or other changes in the indicators folder.
+*   `Assets`: changes to the `src-static` code to generate static assets.
+*   `Testing`: changes to the `test` folder.
+*   `Monitoring`: changes to the `.github` actions files that change how we monitor.
+
+Feel free to propose additional categories, but those need to be reflected here for
+consistency.
+
+### Versioning
```
hannahker, line 69: And here, something like: 

"Update the version in `.signals-version`, following semantic versioning guidelines."

ENVIRONMENT.md
```diff
@@ -0,0 +1,80 @@
...
+
+Data is stored in Azure storage containers, and final output datasets are also
+transferred automatically from Azure into [HDX](https://data.humdata.org). 
+
+### Azure
+
+To read to and from the Azure storage containers, you need the following environment
+variables.
+
+- `DSCI_AZ_ENDPOINT`: URL endpoint to the Azure blob store, with `{glue}` string
```
hannahker, line 16: This one is in the code now, so not needed.

README.md
```diff
@@ -1,28 +1,43 @@
-# CERF Global Monitoring and Alert System (GMAS)
+<img src="https://github.com/OCHA-DAP/hdx-signals/assets/HDXSignalsLogo_V2.png" width=200>
```
hannahker, line 1: This doesn't seem to be loading properly. I think you should be able to reference with markdown syntax like `plot`

CONTRIBUTING.md
```diff
@@ -0,0 +1,98 @@
+# Contributing to hdx-signals
```
hannahker, line 1: Would be useful to also have a section on steps for developer setup. This could be something like: 

1. Clone GitHub repository
2. Install R v 4.3.3
3. Install `renv`
4. Install project dependencies using `renv::init()`
5. Make sure you have the necessary environment variables defined (could be in your `.Renviron` file) -- see `ENVIRONMENT.md` for more details

src/README.html
hannahker, line 1: Why do we have a html file?

src/email/README.md
hannahker, line 1: Can discuss further, but I would actually argue that this level of detail in the tertiary `README`s is more useful directly in the scripts themselves. I fear that it's too easy for this to get out of date.

src/email/components/README.md
hannahker, line 1: Can discuss further, but I'd argue that this level of detail in the docs is going to get out of date too easily. I think it's probably more useful to have this information as comments directly in each script (where it's more likely to be updated when/if someone updates the code).

src/utils/README.html
hannahker, line 1: Also not sure about this html file?

src/README.html
caldwellst: These all need to be `.gitignored.`

README.md
caldwellst: Yeah, it doesn't yet exist until we push to main. Issue is that we don't want it to span the whole page. https://stackoverflow.com/questions/14675913/changing-image-size-in-markdown

hannahker:
Awesome to have these! 😍 Overall, the content makes sense to me and is an excellent improvement. 

A couple high level points for discussion: 

1. Level of detail

Big picture, it would be good to discuss the level of detail that the think is appropriate/necessary for our technical documentation. I'm a bit concerned about creating too much maintenance burden for ourselves by having too much detail. For example, I think that describing the functionality of each script in `README`s is probably not necessary, since we have pretty descriptive script names and good documentation within each script. It's a super slippery slope since as soon as one piece gets out of date, we/others will lose trust in the documentation altogether. It's a tough balance to strike... My feeling is that we're a bit too far into the detailed side of things

3. Getting started with developing

How should new contributors get started with running the system? Which scripts provide an overall entry point? These points took me a while to understand when I started working on this project and so I think would be useful to have explicit guidance around. This should also include a summary of the testing environment variables and how someone would use those to work locally.

zackarno:
yeah i think @hannahker has good points on level of detail as well as making sure the "entry point" for developing/debugging/understanding is clear.

Nonetheless, no objection to merging and iterating over those and other details as we go?

hannahker (approved):
@zackarno agreed! Looks good to merge to me! 💃

PR #163, opened 2024-06-11
init lac ipc labels workaround

### LINKED ISSUE #125: why are we unioning LAC countries?
Was curious why we are combining LAC countries here:

https://github.com/OCHA-DAP/hdx-signals/blob/f5159b9cf8474a615dd8568831f987c68d8e8bb7/src-static/update_iso3_sf.R#L155-L160

Wondering if they should be split like all others?

zackarno:
What do you think about this approach @caldwellst ? Haven't messed with the aesthetics yet!

It's definitely a work-around to avoid substantial refactoring.   

fixes #125 

Relates a bit to the conversation we had here: https://github.com/OCHA-DAP/hdx-signals/issues/74#issuecomment-2123379920 . My experience so far is still that having the admin loading function hardcoded into the geom makes the code harder to extend/refactor.


zackarno:
also if we go with this where are we controlling label text size? Is there a common theme/file somewhere i should tuck it in?

I tried putting `size=8` directly in the `geom_sf_text()` call and it made the text massive

caldwellst:
I don't think we want the full IPC phase classification labels shown. We only want the labels shown for phases that are present in the map. That way the legend clearly shows the range of values.

zackarno:
Okay removed IPC legend issue from this PR and added to a new branch.

I set LAC label size to 3 and think it looks about right - screenshot below is from the generate_signals html output:

image

caldwellst (approved):
Looks good! Just want to update `CHANGES.md` and `.signals-version`.

caldwellst:
Ugh. I lied. Looking at it, maybe we actually want to zoom in just on the analysed areas in this single unique case where we are covering small areas in multiple countries.

zackarno:
but to do the zoom we have to then make all labels and aesthetics (line widths) dynamically sized i think. Do you think we want to go this route?

caldwellst:
Ha, hmm, let's leave it, you're right! Good to merge!

PR #164, opened 2024-06-12
Fix agricultural hotspots

caldwellst:
Re-cloned the repo and new `{renv}` settings seem to update the settings JSON, so that is also included.


hannahker (approved):
LGTM!

Issue #150, opened 2024-06-10
Use `ai_summarizer_without_location` for all indicators for short summaries


caldwellst:
Some of the short summaries are not using `ai_summarizer$ai_summarizer_without_location()`, which checks to ensure the country name is not in the output, and if it is, asks the AI to rewrite the text.

Noticed that in [some campaigns](https://us14.campaign-archive.com/?u=ea3f905d50ea939780139789d&id=f87c74ba37#BRA) the country name is still there, so need to use the above function.

![image](https://github.com/OCHA-DAP/hdx-signals/assets/8086547/d3193c18-7c51-4ce9-abcf-815f4af46cae)

caldwellst:
Fixed by #176

PR #166, opened 2024-06-12
Use the AI summarizer short function to ensure country name removed

caldwellst:
Addresses #150


hannahker (approved):
LGTM! What's the justification for leaving the country name out of the summary?

caldwellst:
Because these only appear in the headlines at the top of the email, so there were instances in the past that looked like:

`Kenya: Decreased rainfall in Kenya poses threat to crop outlook`

PR #168, opened 2024-06-12
Change terminology

### LINKED ISSUE #141: blob terminology
The terminology used in the cloudstorage module is a bit confusing because my understanding is that the `blobs` are the actual files themselves

The hierarchy should be `storage account` -> `container` -> `blob`

https://github.com/OCHA-DAP/hdx-signals/blob/934862b8a6ace590027d3a3e248fdca89842abbd/src/utils/cloud_storage.R#L130-L161

Therefore the `blob_prod()`, `blob_dev()`, and `blob_wfp()` i think should be renamed to `prod_container()`, `dev_container()`, and `wfp_container()`... also `get_blob()` -> `get_container()`

Additionally in functions like `az_file_detect()` i think the `blob` arg should be changed to `container` if that's what it is actually referring to.

if you agree, happy to PR this myself or whatever you think is easier

caldwellst:
Fixes #141

### DIFF
--- a/src/indicators/wfp_market_monitor/utils/raw_market_monitor.R
+++ b/src/indicators/wfp_market_monitor/utils/raw_market_monitor.R
@@ -8,7 +8,7 @@ box::use(cs = ../../../../src/utils/cloud_storage)
 raw <- function() {
   df <- cs$read_az_file(
     name = "BasketCostChange.csv",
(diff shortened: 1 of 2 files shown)


hannahker (approved):
Nice one! Good call @zackarno!

PR #170, opened 2024-06-13
Fix string when no alerts are detected

hannahker:
Today was the first run with no signals detected, which [caused an error](https://github.com/OCHA-DAP/hdx-signals/actions/runs/9497646763/job/26174699811) with the Slack API due to an empty string.


caldwellst (approved):
So weird. Maybe a comment just to explain why signals is `" "` because stands out next to `full_status` above.

PR #171, opened 2024-06-17
Rename `GLOFAS_TEST_RUN` and `test` environment variables

### LINKED ISSUE #94: Clarify `test` and `GMAS_TEST_RUN` variables

hannahker:
Closes #94. These commits were previously in #131, but want to merge this sooner since it's causing frequent conflicts as changes are made to `main`.


src/email/mailchimp/custom_segmentation.R
```diff
@@ -31,18 +31,18 @@ box::use(../../utils/get_env[get_env])
 #'
 #' @param indicator_id Unique indicator ID
 #' @param iso3 Vector of ISO3 codes to generate conditions for
-#' @param test Whether or not this is a test email. If so, only sends to contacts
+#' @param dry_run Whether or not this is a dry run email. If so, only sends to contacts
 #'     with the `hdx-signals-test` tag in Mailchimp.
```
caldwellst, line 35: Should also change this tag in Mailchimp?

src/utils/hs_local.R
```diff
@@ -1,11 +1,11 @@
 #' Checks if this run is a test run
 #'
-#' `gmas_test_run()` looks for the environment variable `GMAS_TEST_RUN`. This is
+#' `hs_local()` looks for the environment variable `HS_LOCAL`. This is
 #' used in other functions so emails are not sent and data not saved back to the
 #' Google Cloud Storage bucket so test runs do not have any impact on the stored
```
caldwellst, line 5: Edit text here to also reflect it's not about test runs (also not about Google Cloud Storage anymore ha), but just that all changes and outputs are local.

caldwellst (approved):
All looks good to me, although hard to know if any have been potentially missed...Will probably find out as we go along. I created a new MailChimp tag `hdx-signals-dry-run` that is tagged to yourself and me for now, can also add @zackarno in if he wants to be pinged in all dry runs! We can then change that `hdx-signals-test` to instead reflect `dry-run` and be more clearly aligned with the env variable.

src/utils/hs_local.R
hannahker: yup good point

src/email/mailchimp/custom_segmentation.R
hannahker: Yeah good point! How would the mechanics of that work?

PR #173, opened 2024-06-18
Update to v2 release of setup action

hannahker:
Have created a new release [here](https://github.com/OCHA-DAP/hdx-signals-actions) so that we don't have to reference the commit from the action.


.github/workflows/test_monitoring.yaml
```diff
@@ -0,0 +1,36 @@
...
+            uses: ocha-dap/hdx-signals-actions@v2
+
+          - name: Test monitoring
+            shell: bash
+            run: |
+              shopt -s globstar
+              set -e
+              for x in test/automated/test_*.R; do Rscript "$x"; done
+            env:
+              HS_TEST: "TRUE"
```
caldwellst: All good, just noting we'll want to change these once you merge the other branch to change these env names, so maybe merge that first then merge this branch!

caldwellst (approved):
All looks good to me! Just a note to wait to merge until you've merged your env var branch, and then reflect those env var changes in the `test_monitoring` YAML file?

.github/workflows/test_monitoring.yaml
hannahker: Yup those should be changed in the other branch!

PR #169, opened 2024-06-12
Add iso3 to shapefiles

caldwellst:
Addresses #139

Issue #93, opened 2024-05-16
Map Viz Issues


zackarno:
Can use this issue to document map visualization as we iterate?

Was just using this took take notes as flipped through the map pngs so is not totally comprehensive and perhaps I can reorganize.

I'm putting the pngs I'm looking at in the drive here: `Predictive Analytics/HDX Signals/tmp/20240516_map_test` so we are looking at the same thing (I guess next time we can set a random seed).

## Random important: 

- [ ] Ukraine png blank

## General

- I think legend placement is good overall! But hard to tell completely as text size is way too small and could be affected by adjustments to extent which are flagged.
- [ ]  City label size seems to be changing with zoom. Lot's are illegible
- [ ] General text sizes seem to change in raw pngs vs those in embedded in email. Maybe we should download raw png from mailchimp and see if those look like the ones here or the ones in email
- [ ] bubble breaks - there are the occasional cases where there is just 1 or 2 bubbles in legend when you'd think there should be 2 or 3 (for example AFG).
- Generally big countries look fantastic but smaller ones less so two factors related to zoom level/extent (_and yes i am making up terms and acronyms_).
  + Small Island Dispersion **(SID)** - small islands spread far apart - hard to map
  + Polygon Simplification **(PS)** - often related to zoom extent, but some appear overly simplified.

## City placement

just flagging, 1. cities that fall outside of polygons, 2. a couple specific examples where text is hard to read (this is an issue that needs to be fixed more generally though).

- [ ] BHR:  city outside of polygon.. Also polygons simplification looks a bit too much
- [ ] HKG: Hong Kong outside of poly
- [ ] ISL: Reykjav.... outside of poly and not text to small to be legible
- [ ] ISR:  city text too small
- [ ] OMN : city text very unreadable
- [ ] PAN hard to read city text

## PS & SID
- [ ] AIA:  PS
- [ ] AND:  PS
- [ ] ASM:  PS & SID
- [ ] ATF: SID
- [ ] ATG: SID (borderline)
- [ ] BES: SID
- [ ] BHR: PS
- [ ] BHS: PS (borderline)
- [ ] BLM: PS
- [ ] BRN: PS
- [ ] BVT: PS
- [ ] CCK: PS
- [ ] COK: PS
- [ ] CXR: PS
- [ ] DNK: PS (borderline)
- [ ] EST: PS (borderline)
- [ ] FLK: PS
- [ ] FRO: PS
- [ ] FSM: PS
[...]

PR #178, opened 2024-06-19
Use complex UN geodata boundaries

caldwellst:
I was able to get access to the privately available complex geo boundaries from the UN, not the simplified ones! I don't want to merge this until we merge #169. But you can test and see what you think, as this fixes the simplified shapes of #93.


zackarno (approved):
code looks good to me! 

I ran `implement_map_test.R` on this branch and put outputs here: 
`Predictive Analytics/HDX Signals/tmp/20240619_complex_bounds`

haven't done a thorough comparison against previous files to see what % have improved, but I believe it to be an improvement nonetheless

caldwellst:
Succeeded in updating static assets using this approach: https://github.com/OCHA-DAP/hdx-signals/actions/runs/9594645807

Issue #142, opened 2024-06-06
Explore `{box.linting}` for linting


caldwellst:
Was made aware of a package extension to `{lintr}` to lint box modules, [{box.linters}](https://cran.r-project.org/web/packages/box.linters/index.html). Should definitely explore usage, as out of the box should fix things like us having to disable the `OBJECT_USAGE_LINTER` in `{lintr}`, which means we aren't catching extraneous or missing imports, failing to catch `library::function()` calls, etc.

caldwellst:
Have solved the vast majority of issues. I can successfully lint on my machine, yet the linting is failing on remote. See the latest failure here:

https://github.com/OCHA-DAP/hdx-signals/actions/runs/9746045472

```
ERROR [2024-07-01 14:28:10] Environment variable 'DSCI_AZ_SAS_PROD' is empty or not set.

Error: Error: Linter 'box_unused_attached_mod_linter' failed in /home/runner/work/hdx-signals/hdx-signals/src-static/update_acled_info.R: 
Execution halted
Error: Process completed with exit code 1.
```

So, the failure is on the first script: https://github.com/OCHA-DAP/hdx-signals/blob/feature/box-lint/src-static/update_acled_info.R

This script imports a module `cs = src/utils/cloud_storage`. Inside that module, it uses `get_env$get_env()`: https://github.com/OCHA-DAP/hdx-signals/blob/736e8c71c9bd31d441308b11ee1d64f533f15870/src/utils/cloud_storage.R#L208

`get_env()` retrieves an environment variable, but throws an error if it doesn't exist. This is the error we see in the lint run. https://github.com/OCHA-DAP/hdx-signals/blob/feature/box-lint/src/utils/get_env.R

However, when using the default linters in `lintr`, we were receiving no such errors. I can remove environment variables from my machine and recreate the error locally, so it is not solely present on remote runs.

I have tried to create various minimal file directories and files to test linting and create an issue on the box.linters repo. However, I cannot reproduce the error with any simple code examples, which leads me to believe there is something particularly unique going in this complex series of calls that is generating the error.

caldwellst:
Closed by #181

Issue #123, opened 2024-05-28
possible fix for `unable_to_load_module x in path y`


zackarno:
I get this type of error every time I restart my R-session and first try to load a custom module. After the first error, it tends to go away.

```
Error in box::use(./module_name) : 
  unable to load module “./module_name”; not found in “/directory/path”
```

@caldwellst @hannahker  - do you get this also? I guess the GH runner doesnt!?

@caldwellst I remember we talked about this somewhere , but don't think we came to a clear solution and I wasn't able easily find the discussion.

Anways, the  [{box} FAQ page](https://klmr.me/box/articles/faq.html) describes a very similar, but slightly different (I think) error message and the solution proposed it just to run this:

```
box::use(r/core[...])
```

This does stop the error from occurring on my computer as well. It seems kind of annoying to put at the top of every script though. If this is in fact a good solution, I wonder if it could possibly make sense/work to have some sort of onLoad() type deal for it .

caldwellst:
That's not what using the core functions does. `box::use(r/core[...])` solves the error you find if you try to run functions like `View()` or `read.csv()`, because they are not attached by default in `{box}`. So, you have to do:

```r
box::use(utils)
utils$View()
utils$read.csv()`
```

The error you describe above is one that I get when I am interactively running line by line through a script in RStudio into my R console. It seems to be an error with not recognising the relevant pathing, but has nothing to do with the solution above, hence why there are never errors if you run the scripts from your terminal or on GitHub Actions. I've tried to look for this issue on the `{box}` repo before, but haven't found anything. Will have a look again and post something on there.

caldwellst:
https://github.com/klmr/box/issues/365

zackarno:
yea interesting, but  I wonder why `box::use(r/core[...])` does fix it!? I wonder if one of those non-base, but core packages that are typically loaded in R-sessions somehow help?

glad you opened an issue though - that should help!

Two additional things I noticed:

- Don't think it has to do w/ `command` + `enter` workflow. Happens if you just copy paste into the console.
- Does not happen if you are working in script saved to root `.rproj` directory.

caldwellst:
Closing, and we can hopefully get some information in https://github.com/klmr/box/issues/365. However, may be an issue outside of the control of `{box}`, remains to be seen what the maintainer of `{box}` thinks.

PR #181, opened 2024-06-20
Move to box.linters

caldwellst:
Addresses #142. Currently stuck because it seems `{box.linters}` does not follow module relevant pathing if all imports are not directly nested under the file calling `box::use()`. That is, `box::use(../path/to/module)` will fail. Opened up issue [here](https://github.com/Appsilon/box.linters/issues/110).


caldwellst:
Have followed the advice in https://github.com/Appsilon/box.linters/issues/110, which is to move to use directory level pathing in a way that is very nice. It will help solve #123 and mean we can run `box::use()` in the console directly. However, now, the lint is flagging up EVERY SINGLE FUNCTION. So something must be going on here that needs investigation.

caldwellst:
Create a reprex for the latest issue faced in the `{box.linters}` repo: https://github.com/Appsilon/box.linters/issues/111

PR #185, opened 2024-06-24
Create data dictionary

caldwellst:
Add script to create data dictionary that will be added to HDX. I have also added a workflow to update this, and did some slight updates to our other workflows to split out the update of static assets. We update all the locations data and metadata together in one, ending with the metadata file that is pushed to HDX. From there, the rest can be updated separately as necessary, including the new `update_data_dictionary.yml` flow.


.github/workflows/update_acled_info.yml
```diff
@@ -0,0 +1,31 @@
...
+          HS_LOCAL: ${{ inputs.HS_LOCAL || 'FALSE' }}
+          HS_HDX_BEARER: ${{ secrets.HS_HDX_BEARER }}
+          DSCI_AZ_SAS_DEV: ${{ secrets.DSCI_AZ_SAS_DEV }}
+          DSCI_AZ_SAS_PROD: ${{ secrets.DSCI_AZ_SAS_PROD }}
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v3
+
+            - name: Set up environment
+              uses: ocha-dap/hdx-signals-actions@2dfcc5c8796903a01b393b8ebed26d2582f305ff
```
hannahker, line 27: Can update this now to `actions@v2`, since I've created a release in that repo.

.github/workflows/update_data_dictionary.yml
```diff
@@ -0,0 +1,31 @@
...
+          HS_LOCAL: ${{ inputs.HS_LOCAL || 'FALSE' }}
+          HS_HDX_BEARER: ${{ secrets.HS_HDX_BEARER }}
+          DSCI_AZ_SAS_DEV: ${{ secrets.DSCI_AZ_SAS_DEV }}
+          DSCI_AZ_SAS_PROD: ${{ secrets.DSCI_AZ_SAS_PROD }}
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v3
+
+            - name: Set up environment
+              uses: ocha-dap/hdx-signals-actions@2dfcc5c8796903a01b393b8ebed26d2582f305ff
```
hannahker, line 27: Same as above!

.github/workflows/update_indicator_mapping.yml
```diff
@@ -0,0 +1,31 @@
...
+          HS_LOCAL: ${{ inputs.HS_LOCAL || 'FALSE' }}
+          HS_HDX_BEARER: ${{ secrets.HS_HDX_BEARER }}
+          DSCI_AZ_SAS_DEV: ${{ secrets.DSCI_AZ_SAS_DEV }}
+          DSCI_AZ_SAS_PROD: ${{ secrets.DSCI_AZ_SAS_PROD }}
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v3
+
+            - name: Set up environment
+              uses: ocha-dap/hdx-signals-actions@2dfcc5c8796903a01b393b8ebed26d2582f305ff
```
hannahker, line 27: And here

caldwellst:
Yeah, previously was controlling the order, but actually, most of them are unrelated. The only ones that are critically dependent on each other are in `update_locations_data.R`, which is the ones that require first capturing how many locations we cover and then for each location, doing various calculations, finalising that with the overall locations metadata file.

The data dictionary file, ACLED info, and indicator mapping are not at all related, so figured do we need to do all the computation for the locations, like creating the ADM0 files, calculating centroids, etc. if we aren't updating them? Happy to put back all in one file but that was my thought process when I was creating the dictionary file because I was thinking, if we want to update the dictionary file, we would have to wait a long while for everything to update, even if there were no dependencies.

hannahker (approved):
LGTM! 💃

PR #192, opened 2024-06-28
Feature/wfp changes

caldwellst:
Addresses requests for changes from WFP. Primarily, it adds a lot of caveats to the dataset. Because we include the partner logo, we have instead added a set of text about the dataset monitored to the bottom of the email. This has been added for other datasets as well.


caldwellst:
Multiple changes have come about in conversation with the WFP team, and you can see the test campaign I've shared with them in those mailings. However, I think all substantial requests have been dealt with, so limited changes from here on out.

hannahker:
Would you be able to share the link to the updated campaign? I'm just getting an error with the one from the email thread.

caldwellst:
From the latest email reply to Angela: https://us14.campaign-archive.com/?u=ea3f905d50ea939780139789d&id=650b09e3d3

src/email/components/email_body.R
```diff
@@ -32,53 +32,43 @@ box::use(./line_block)
...
-    map_url = "",
-    plot2_title = "",
-    plot2_url = "",
-    other_images_urls = "",
-    other_images_captions = "",
-    summary_long = "",
-    summary_short = "",
-    summary_source = "",
-    further_information = "",
+    campaign_details,
```
hannahker, line 35: Nice simplification here!

hannahker (approved):
Great. Changes look good to me!

Issue #101, opened 2024-05-20
Save backup of Mailchimp subscriber base to Azure


caldwellst:
The Mailchimp servers have gone down today, so unable to use the API reliably. Obviously, in a worst case scenario we would lose our subscriber base to the system if their data was wiped or the service went down permanently. Everything else is reconstructible quite easily, it is only the subscribers that are a high risk dataset stored on the Mailchimp servers.

Although this is highly unlikely, would be good to have a backup script in `src-static` that is run every time we update static files, but also running on a CRON job every week or so to create a subscriber backup. It would be very simple, as `audience$mc_members()` already extracts members in a list from the JSON format and we could just save as a JSON file directly to Azure.

PR #196, opened 2024-07-02
Implement Mailchimp backup

caldwellst:
Addresses #101.


hannahker (changes requested):
Looks good! I would just suggest putting the script under `src/email/mailchimp`, since I think that directory is a good fit and we don't need to add any new directories.

caldwellst:
Alright, moved under `email/mailchimp`!

README.md
```diff
@@ -14,6 +14,7 @@ documentation for details on datasets included and how signals are generated.
 
 The code is structured as below. Refer to the directories for specific details.
 
+- runs: Scripts for actual runs, called in `.github` workflows.
```
hannahker, line 17: Can remove this now!

hannahker (approved):
Looks great! Just see the comment to update the `README`, then should be good to merge.

README.md
caldwellst: Doh! Good catch.

Issue #180, opened 2024-06-20
Add tick marks to represent the exact position on the x-axis


caldwellst:
A piece of feedback I received from Yakubu from CERF:

> I hope that you are doing good. Thanks once again for the great work on signals. One quick question: could you please look into the possibility of adding tick marks to represent the exact position of the years on the x-axis if it makes sense? 

I think that makes sense for our time series plots! What do y'all think @hannahker @zackarno?

zackarno:
Yea seems like a good point

caldwellst:
Okay, I think we want it for [ACLED](
https://mcusercontent.com/ea3f905d50ea939780139789d/images/1c2752d2-a10e-180f-c74d-ee3226ea8f47.png
), [IDMC](
https://mcusercontent.com/ea3f905d50ea939780139789d/images/815fd87f-c411-fa2f-a293-ea2d660ab16d.png
), and [IPC](
https://mcusercontent.com/ea3f905d50ea939780139789d/images/9c0a48f4-afbb-affb-57a1-254702e700f1.png
).

I don't think it will work for [WFP](https://mcusercontent.com/ea3f905d50ea939780139789d/images/c712830c-1fe6-2feb-2f12-f414fe2199c8.png), and isn't necessary for [JRC](https://mcusercontent.com/ea3f905d50ea939780139789d/images/5a3b69fc-4a11-0f42-cf68-dc9f63ddc138.png).

caldwellst:
Closed by #199

PR #199, opened 2024-07-03
Add ticks to plot_ts

caldwellst:
Addresses #180. Implemented for `plot_ts()`, so will be used for IDMC, ACLED, and WHO:

image

Implemented in `plot_food_insecurity` as well:

image


src/indicators/ipc_food_insecurity/utils/plot_food_insecurity.R
```diff
@@ -177,7 +179,10 @@ food_insecurity_ts <- function(df_wrangled, df_raw, title, date) {
     gg$expand_limits(
       y = 0
     ) +
-    theme_signals$theme_signals("subtitle") +
+    theme_signals$theme_signals(
+      margin_location = "subtitle",
+      x_axis_ticks = TRUE
```
hannahker, line 184: Wouldn't this arg need to be added to IDMC, ADLED, and WHO plotting functions as well?

caldwellst: No, because within those functions, they are calling `plot_ts()` to create the generic line plots, so we just have to adjust it in `plot_ts()`. See the `plot_displacement.R` script for example: 

https://github.com/OCHA-DAP/hdx-signals/blob/9e2a6d8e5d2928bfd6c8325fdd73ca28cdf61cf0/src/indicators/idmc_displacement/utils/plot_displacement.R#L63

zackarno (approved):
LGTM! 

1 comment you can address here or elsewhere (which ever is more convenient):
I was thinking it might be nice to make the displacement and food security time series plots have yearly breaks/ticks on the x-axis (like armed conflict). From the last several emails it looks like there should be space.

src/images/plots/breaks_date.R
```diff
@@ -0,0 +1,24 @@
...
+#' @export
+breaks_date <- function(x) {
+  max_year <- lubridate$year(max(x))
+  min_year <- lubridate$year(min(x))
+  n_years <- max_year - min_year
+  if (n_years >= 2) {
+    scales$breaks_width(width = "1 year")(x)
+  } else {
+    scales$breaks_pretty()(x)
+  }
```
zackarno, line 23: 1. in Roxy you say if 3 or more years, but in code its `n_years > = 2`.
2. I was actually thinking the logic would be opposite ... i.e if there were 10 years we'd want breaks_pretty() to do it's thing so it didn't get busy.
3. What does breaks_pretty() do if 1 year or less? it just cuts the months nicely?

caldwellst: Yeah, it's actually no longer `n_years`, it's really calculating `years_diff`. I think overall can change this to be using the year breaks if 3 or more years covered up to something like 10 years. But your suggestion for the plot in Slack for instance covers 8 years, so I think we still want yearly coverage up to then or so.

`breaks_pretty()` chops up months or quarters depending on the data.

PR #203, opened 2024-07-09
Add scalar condition

caldwellst:
Pushed quickly like a dingle before the lint finished. Now have fixed the lint by ensuring I use the scalar logical operator.


hannahker (approved):
😆

Issue #204, opened 2024-07-10
Add functionality to archive alerts in the rare case where we don't want to send them


caldwellst:
With WFP market monitoring, it has been requested not to send the alert for the previous monitor, but we don't want to just delete it. Previously we only archived in `first_run`. So will need to add functionality to archive in `triage_signals()`.

PR #205, opened 2024-07-10
Implements #204

caldwellst:
Adds ability to archive signals when triaging. So, the email campaigns are removed, but the data is still added to `output/signals.parquet`. Currently necessary for WFP market monitor as they requested not to send the previous bulletin. Likely won't use this functionality much at all, but is nice to have.

Can run through the functionality to test doing:

```r
triage_signals$triage_signals("wfp_market_monitor")
```


hannahker:
Closes #204

hannahker (approved):
Nice addition! Looks good to me!

PR #211, opened 2024-07-16
Create first run and dry run env variables to better track throughout generate_signals

caldwellst:
Implemented the setup where first run and dry run are env variables. This was used to better track their status through `generate_signals()` and its internals without having to constantly pass the variables through. This was necessitated because `create_images()` had a small issue that meant that when generating the historic first runs, first runs within the past 90 days were not properly filtered to the campaign date. The only simple fix was to implement the environment variable `HS_FIRST_RUN`, otherwise it would've been a complex web of changes to bring the `first_run` parameter to all of the `plot_{indicator}` modules and all the way to `create_images()`.


caldwellst:
Succeeded in running these workflows for testing:

- [WFP](https://github.com/OCHA-DAP/hdx-signals/actions/runs/9958635617)
- [ACLED](https://github.com/OCHA-DAP/hdx-signals/actions/runs/9958524241)
- [IPC](https://github.com/OCHA-DAP/hdx-signals/actions/runs/9958263639)
- [IDMC](https://github.com/OCHA-DAP/hdx-signals/actions/runs/9958260439)

ENVIRONMENT.md
```diff
@@ -5,6 +5,28 @@ the maintainers if you do not have them and feel you should be given access. The
...
+
+Environment variables are used to determine how `generate_signals()` works. These
+are all logical variables that are read internally to determine how the underlying
+functions behave and the final output returned. The values should be `TRUE` or `FALSE`.
+
+- `HS_LOCAL`: if `TRUE`, prevents functions from writing to the Azure containers,
+sending emails with Mailchimp, or calling the OpenAI endpoint. All other functionality
+occurs, but the outputs thus remain local to the running machine. Useful for testing
+and interactive development.
+- `HS_DRY_RUN`: if `TRUE`, `generate_signals()` does a dry run, which ensures
```
hannahker, line 18: Would clarify that in the dry run, you're guaranteed to have signals generated for testing.

caldwellst: https://github.com/OCHA-DAP/hdx-signals/pull/211/commits/95cff7c72893d8052891ae23a3e630c2f39253f0

hannahker (commented):
Overall these changes make sense, but have some thoughts to try and improve the simplicity/readability. 

I'm not sure if this is the convention in R, but I don't love the set up of defining the variable inside a function and then importing and calling that function throughout the code base. IMO, something like `dry_run <- hs_dry_run$hs_dry_run()` is quite verbose and not easy to understand what's going on. I think it's also a bit redundant to have separate `hs_dry_run.R` and `hs_first_run.R` files with very short functions. 

Could we consolidate the `utils/` functions to have a single `globals.R` script that has all of our environment variables? Is it also possible to just import variables throughout the codebase instead of functions?

PR #216, opened 2024-07-17
Add another custom location name handler to location_codes

caldwellst:
Addresses error seen here with new JRC data update: https://github.com/OCHA-DAP/hdx-signals/actions/runs/9970458118

They changed names, and had an error in the names that they used, so it broke the script. Would be safer to use their IDs, so implemented a mapping from ASAP IDs to ISO3 codes based on files that they sent this week to Tristan.


src-static/README.md
```diff
@@ -19,6 +19,10 @@ website. Updated in `update_centroids_sf.R`.
 `update_cities_sf.R`.
 - `acled_info.parquet`: Data on the start date of ACLED coverage for locations
 and the web URL to the location page. Updated in `update_acled_info.R`.
+- `asap_iso3.parquet`: ASAP0 to ISO3 mapping file. Used in `location_codes.R`.
+- `gaul1_asap_FAO_Mapping_v8.csv`: CSV file provided directly by JRC to map
+ASAP0 IDs to GAUL codes, which can then map to ISO3. Used to created
```
hannahker, line 24: Just a minor typo: `created` vs `create`

hannahker (approved):
Looks good!

src-static/README.md
caldwellst: https://github.com/OCHA-DAP/hdx-signals/pull/216/commits/6355c442d0b1eed309bad20d41b9b00d0af4ef56

PR #219, opened 2024-07-22
Bring click tracking back

caldwellst:
Since the quarantine has been resolved, we are going to try and re-enable click tracking for new campaigns.


zackarno (approved):
LGTM! I'm not sure how to actually check it, but can't see how it possibly wouldn't work and I assume you are correct on the quarantine

Issue #83, opened 2024-05-14
Remove warning generated from `geom_sf()`


caldwellst:
Somehow have maps not generate warnings when plotting longitude/latitude in `geom_sf()`.

zackarno:
will look into it more... just putting down a couple notes from my first glance:

A couple notes:

## Warning message:

```
Warning message:
In st_point_on_surface.sfc(sf::st_zm(x)) :
  st_point_on_surface may not give correct results for longitude/latitude data
```

As indicated the warning message originates from `sf::st_point_on_surface()`, however in our context warning is only generated when labeling the cities inside`geom_cities()` with `geom_text_repel()`. Interestingly no warning when plotting the point(s) themselves inside `geom_cities()`

```
ggrepel$geom_text_repel(
        data = sf_cities,
        mapping = gg$aes(
          label = name,
          geometry = geometry
        ),
        stat = "sf_coordinates",
        min.segment.length = Inf,
        family = "Source Sans 3",
        color = "black"
      )
```

zackarno:
I think i found a work-around -- which is to add fun.geometry argument. 

```
ggrepel$geom_text_repel(
        data = sf_cities,
        mapping = gg$aes(
          label = name,
          geometry = geometry
        ),
        stat = "sf_coordinates",
        fun.geometry = \(x) sf::st_centroid(x , of_largest_polygon = F),
        min.segment.length = Inf,
        family = "Source Sans 3",
        color = "black"
      )
```

Some points I don't fully understand yet:
1. if `of_large_polygon=T` it works, but the point moves from default text_repel location. Not necessarily in a bad way
2. running `st_centroid()` directly on the data input does not remove the warning message

I got the `fun.geometry` from: 
https://stackoverflow.com/questions/73623865/ggplot2-and-eurostat-row-removed-due-to-missing-values-geom-text-using-geom-s

So there might be some more interesting tidbits/clues there.

caldwellst:
Closed by #75

PR #220, opened 2024-07-30
Suppress warnings in map_food_insecurity

caldwellst:


hannahker (approved):
LGTM!

zackarno:
the suppressWarnings on `fun.geometery` works!  Also another solution was found in #83 I believe

zackarno:
the suppressWarnings on `fun.geometery` works!  Also another solution was found in #83 I believe

caldwellst:
#83 didn't resolve the issue. It resolved the issue in the linked SO post because calculating the centroid explicitly removed a specific warning, but the specific warning of plotting coordinates was still there because `st_centroid()` still returned coordinates. So now for all of those point plotting I am using `suppressWarnings()` around the `fun.geometry`.

PR #221, opened 2024-07-31
Add functionality to reduce basemap if necessary

caldwellst:
We ran into our first instance where the basemap for a country needed reduction, as the scale did not look good. This was for the USA where by default the contiguous USA and Alaska are plotted. We now have the functionality to check if spatial data falls within Alaska and if not, Alaska is dropped from the base map.

This will be useful for other countries where we drop distant territories or islands from their base spatial data. They are currently just dropped permanently in `update_adm0_sf.R`. If we ever need to use those territories, we can add them back to the basemap and then handle those territories `geom_adm0.R` instead.


caldwellst:
Successful test runs here:

- [IDMC](https://github.com/OCHA-DAP/hdx-signals/actions/runs/10178656131)
- [IPC](https://github.com/OCHA-DAP/hdx-signals/actions/runs/10178563332)
- [ACLED](https://github.com/OCHA-DAP/hdx-signals/actions/runs/10178459988)

src-static/update_cities_sf.R
```diff
@@ -39,7 +39,7 @@ update_cities_sf <- function(iso3) {
       pop_sf,
       adm0_a3 == iso3,
       !(name %in% c("Laayoune", "Monaco", "Singapore", "Vatican City", "Manama",
-                    "Hong Kong", "Jerusalem"))
+                    "Hong Kong", "Jerusalem", "Atlanta", "Los Angeles", "Chicago"))
```
hannahker, line 42: What's up with these cities specifically?

caldwellst: There were too many across the United States, so the map was getting cluttered. Removed those that were near other cities (kept Miami instead of Atlanta, San Francisco instead of Los Angeles, and Denver instead of Chicago).

image

hannahker: Gotcha! Makes sense!

src/images/maps/gg_map.R
```diff
@@ -13,8 +13,8 @@ box::use(../plots/theme_signals)
 #' @returns ggplot object
 #'
 #' @export
-gg_map <- function(iso3) {
+gg_map <- function(iso3, ...) {
```
hannahker, line 16: Curious, is the `...` an R convention (like Python's `kwargs`)?

hannahker (approved):

src/images/maps/gg_map.R
caldwellst: Yeah! Pretty similar. Although it can also accept positional argument arguments, not just named arguments. So you can do something like `a <- function(...) sum(...)` and just pass everything on. But named arguments are the typical use, where you are wrapping one function that you want users to have the option to pass arguments to using the ellipsis!

PR #222, opened 2024-07-31
Filter time series to stop at date of signal

caldwellst:
Filters time series to stop at date of signal. This ensures that plots clearly display the data that generated the signal, as previously, filtering to latest available date meant that the aggregated data generating the signal might not be used in the plot (since we are daily checking a 30 day rolling sum).


hannahker (approved):
Nice! I like this update :)

PR #223, opened 2024-07-31
Monitor all event types in acled_conflict

caldwellst:
We are currently monitoring only 3 violent event types for ACLED conflict: battles, explosions/remote violence, and violence against civilians. There is ongoing violence in Bangladesh that we currently would not generate a signal for, because it is fatalities against protestors. This PR would change our system to monitor all ACLED event types and capture this Bangladesh event if we decide to go forward with it.


caldwellst:
@hannahker Can we try and get this merged today, that way it's available for the ACLED run tomorrow morning European time? I can merge later tonight unless substantial changes necessary!

.github/workflows/publish_prompts.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             run: |
               git config user.email "gha@email.com"
               git config user.name "GHA"
-              git diff --quiet || (git add add docs/PROMPTS.md && git commit -m "Publish PROMPTS.md")
```
hannahker, line 46: Nice catch

hannahker (approved):
All looks good to me from a technical perspective!

src/indicators/acled_conflict/utils/map_conflict.R
```diff
@@ -24,7 +24,7 @@ map <- function(df_alerts, df_wrangled, df_raw, preview = FALSE) {
   df_map <- df_alerts |>
     dplyr$mutate(
       title = paste0(
-        "Conflict events since ",
+        "Reported events since ",
```
hannahker, line 27: Would they not still be conflict events?

caldwellst: Putting as Reported events because they potentially cover just protests, riots, or other non-violent events in the dataset (although we only aggregate fatalities).

Issue #133, opened 2024-05-30
Configure logger in profile


caldwellst:
I see that in a lot of scripts, we are running `hs_logger$configure_logger()`. It also seems not to be parameterised. Can we just place it straight in the `.Rprofile` and remove it elsewhere?

hannahker:
Closed with #224

PR #224, opened 2024-08-01
Move `logger` setup to `.Rprofile`

hannahker:
Addresses #133

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,5 +1,9 @@
+box::use(src/utils/hs_logger)
+
 source("renv/activate.R")
 
 if (!interactive()) {
   options(warn = 2)
 }
+
+hs_logger$configure_logger()
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,9 +1,9 @@
-box::use(src/utils/hs_logger)
-
 source("renv/activate.R")
+box::use(logger[log_threshold])
 
 if (!interactive()) {
   options(warn = 2)
 }
 
-hs_logger$configure_logger()
+log_level <- Sys.getenv("LOG_LEVEL", unset = "INFO")
+log_threshold(log_level)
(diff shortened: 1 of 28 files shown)


hannahker:
Hey @caldwellst do you have any idea what might be going on with the `renv` setup in these actions? I'm having a hard time debugging here.

caldwellst:
Aha yes, I see! The errors are coming about because `.Rprofile` is loaded at the beginning of any R session. However, in this case, it's loading at the beginning of an R session where we haven't restored our installed libraries yet with `{renv}` (hence why `{box}` isn't available.

I think best would be to wrap the call to `{box}` to check if it's been installed, then run the relevant code block. We can assume if `{box}` installed then also `{logger}` and other necessary packages are too, I think.

Something like:

```r
if ("box" %in% installed.packages()) {
  ...
}
```

.Rprofile
```diff
@@ -1,5 +1,9 @@
 source("renv/activate.R")
+box::use(logger[log_threshold])
```
caldwellst, line 2: Yeah, just move all of this and other new code into the block checking if `{box}` installed.

src/utils/hs_logger.R
```diff
@@ -1,12 +1,4 @@
-box::use(logger[log_threshold, log_info, log_debug])
-
-#' Sets up the logging environment
-#'
-#' @export
-configure_logger <- function() {
-  log_level <- Sys.getenv("LOG_LEVEL", unset = "INFO")
-  log_threshold(log_level)
-}
+box::use(logger[log_info, log_debug])
```
caldwellst, line 1: Just while we're here, I noticed in `monitoring_log_setup()` that `FIRST_RUN` should be `HS_FIRST_RUN`, so maybe can just put that fix into this PR.

caldwellst (approved):
Looks good to me!

PR #229, opened 2024-08-13
Increase buffer when checking overlap between points and admin boundaries

caldwellst:
An ACLED alert today was due to go out when two points were detected that were not covered by the administrative boundaries. One of those was actually miscoded, which we've alerted ACLED about, so we want the system to run a bit.

However, another was an event at sea in the gulf of Guinea. This was very close to the coast, and the only one we've ever encountered, so extending the buffer to capture these. However, in the long run might need to consider what to do about that, as it seems it happens on occassion in other conflicts (for instance, Israel there are 7 instances historically). Will add separate issue about a long term discussion.


src/images/maps/geom_adm0.R
```diff
@@ -56,15 +56,16 @@ geom_adm0 <- function(iso3, ...) {
...
 #' @param x_list All geometries to check
 #' @param y sf class POLYGON/MULTIPOLYGON to use to check if all x falls within
-#' @param dist distance in meters (default = 1000)
 #'
 #' @return TRUE if all of `x_list` fall within `y`.
 assert_covered_all <- function(x_list, y) {
   y_buff <- suppressWarnings(
     suppressMessages(
-      sf$st_buffer(x = y, dist = 0.1)
+      sf$st_buffer(x = y, dist = 0.3)
```
hannahker, line 68: Would this create additional historical alerts if we regenerated all of the old signals?

caldwellst: No, because if something isn't covered, an error is generated, so we wouldn't have generated historical alerts if we had this problem before. Seems to be something we've only run into now with this naval battle. There have been other naval battles in the ACLED data but must be quite rare and not been included in a Signal yet (or only been within 0.1 degrees).

caldwellst: Oh and to clarify, checking is done at the time we generate the map, which is after creating the signals. We do no filtering on the data for the signals, hence why we check later that the map makes sense.

hannahker: Ok gotcha! That makes sense.

hannahker (approved):
LGTM!

PR #231, opened 2024-08-14
Change Signals infrastructure to only monitor past 60 days

caldwellst:
Was decided based on review of an ACLED alert that was nearly 90 days old that this was too long ago, so we now only look for new signals in the past 60 days.


hannahker (approved):
LGTM!

PR #232, opened 2024-08-19
Fix audience call to return more than 1,000 members if necessary

caldwellst:
Last week we did some checks of the main OCHA CHD mailing list (30k+ subscribers compared to our 500+) and I noticed the `mc_members()` function hadn't been properly set up to iterate and make all calls necessary to get the full members list. Have made a small fix so that if we go over 1,000 subscribers this won't break.


hannahker (approved):
Good fix!

PR #233, opened 2024-08-19
Ensure triage_signals() user prompt is not cut off

caldwellst:
When we added the `ARCHIVE` option to `triage_signals()`, we ran up against a 256 character limit in `readline()` when asking for user prompts. So I've moved the user prompt into a `cat()` call to print that to the console, and then ask for user input with `readline()` following.


hannahker (approved):
Also nice catch!

Issue #243, opened 2024-10-09
Refinement of `triage_signals` Function: Command Handling, Logging, and Confirmation


martinig94:
The `triage_signals` function could benefit from some improvements to enhance its robustness, logging, and user interaction.

**Suggested Improvements:**
- Command Validation: Introduce a DO NOTHING command to exit from the process.
- If an invalid command is provided, the system should recursively ask the user for a valid action instead of proceeding with no action or failing silently.

**Enhanced Logging and console output:**

Add more detailed logging to capture key events and actions within the function. This will help with debugging and provide better context for decisions made during the process. Set proper priority level to be sure to have the needed info as console output

**Double Confirmation for Critical Actions:**

Implement a double confirmation step before performing irreversible actions like approval, deletion, or archiving. This will reduce the risk of accidental actions by requiring explicit user confirmation.

**Why is this important?**
These enhancements would reduce potential errors, provide better visibility into the function’s operations, and make the system more user-friendly by handling incorrect inputs more effectively.

**Impact:**

Improve error handling and system stability.
Provide clearer feedback to users.
Ensure that critical operations are only executed after explicit confirmation.

caldwellst:
Closed by #246

PR #244, opened 2024-10-11
Add confirmation for DELETE and APPROVAL, add DO_NOTHING action 

martinig94:
Deals with with issue #243

### DIFF
--- a/src/signals/triage_signals.R
+++ b/src/signals/triage_signals.R
@@ -30,8 +30,13 @@ box::use(
 #' delete all campaign content from Mailchimp and then delete all of the rows
 #' from `output/{indicator_id}/signals.parquet`.
 #'
-#' - `Any other user input`: Do nothing, in which you can decide later manually what to do, and use
+#' - `DO_NOTHING` Do nothing, in which you can decide later manually what to do, and use
 #' `delete_campaign_content()` or `send_signals()` yourself manually.
+#'
+#' - `Any other user input`: the user will be asked again what's the desired action to be taken.
+#'
+#' The`APPROVE` and `DELETE` command will require a double confirmation given the criticality of the action.
+#' To confirm the user needs to type `I CONFIRM`
 #'
 #' There is no way to completely delete the alerts, which would essentially
 #' determine that we do not want to send an alert at all. This would mean changing
@@ -140,8 +145,9 @@ preview_campaign_urls <- function(campaign_urls) {
 #' @param df Signals data frame
 #' @param fn_signals File name to the signals data
 #' @param test Whether or not the signals were in the test folder.
-approve_signals <- function(df, fn_signals, test) {
+approve_signals <- function(df, fn_signals, test, user_command='init'){
   user_name <- get_env$get_env("HS_ADMIN_NAME")
+  while (!(user_command %in% c('APPROVE', 'DELETE', 'ARCHIVE', 'DO_NOTHING'))){
   # send message to user
   cat(
     "Tell us what you want to do with the following commands:\n\n",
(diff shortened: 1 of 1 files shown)


src/signals/triage_signals.R
```diff
@@ -140,8 +145,9 @@ preview_campaign_urls <- function(campaign_urls) {
 #' @param df Signals data frame
 #' @param fn_signals File name to the signals data
 #' @param test Whether or not the signals were in the test folder.
-approve_signals <- function(df, fn_signals, test) {
+approve_signals <- function(df, fn_signals, test, user_command='init'){
```
caldwellst, line 148: Make sure to check the linting for style matching. You can run it locally just running `lintr::lint_dir()`. All of the settings we use are in `.lintr`. For instance, always need spaces around arguments, and no single quotes (unless necessary for wrapping double quotes in a string for instance). So, `user_command = "init"`.

src/signals/triage_signals.R
```diff
@@ -150,11 +156,27 @@ approve_signals <- function(df, fn_signals, test) {
...
     "file so you can recreate later.\n",
     "ARCHIVE: Delete the email campaign, but move the alert to `output/signals.parquet`\n",
-    "Any other input: Do nothing, so you can decide later.\n",
+    "DO_NOTHING: Do nothing, so you can decide later.\n",
     sep = ""
   )
 
-  user_command <- readline(prompt = "Your command: ")
+  user_command <- readline(prompt = "Your command: ")}
+  if (user_command %in% c("APPROVE", "DELETE")) {
```
caldwellst, line 164: I would also add the confirmation for archiving. It's not as critical but it does delete the draft campaign that would go out to users and only maintains the internal campaign.

src/signals/triage_signals.R
```diff
@@ -150,11 +156,27 @@ approve_signals <- function(df, fn_signals, test) {
...
+    # send message to user
+    cat(
+      "You typed the command: ", user_command, "\n",
+      "Given the criticality of the action please type I CONFIRM to proceed with the action selected",
+      sep = ""
+    )
+
+    user_command_confirmation <- readline(prompt = "Your command: ")
+    if (user_command_confirmation != "I CONFIRM"){
+      stop(
```
caldwellst, line 174: Use `call. = FALSE` so that the error message appears cleaner to the user. Basically, it means, it won't show the traceback, which they don't necessarily need here, it's more about what they type as input.

src/signals/triage_signals.R
```diff
@@ -150,11 +156,27 @@ approve_signals <- function(df, fn_signals, test) {
...
+        "The process was not confirmed and it was aborted."
```
caldwellst, line 175: Maybe something slightly clearer here? Since we know the `user_command`, could say something like: 

```
glue$glue("The process was not confirmed, you will need to re-run `triage_signals()` to {user_command} the signal.")
```

src/signals/triage_signals.R
```diff
@@ -30,9 +30,14 @@ box::use(
 #' delete all campaign content from Mailchimp and then delete all of the rows
 #' from `output/{indicator_id}/signals.parquet`.
 #'
-#' - `Any other user input`: Do nothing, in which you can decide later manually what to do, and use
+#' - `DO_NOTHING` Do nothing, in which you can decide later manually what to do, and use
 #' `delete_campaign_content()` or `send_signals()` yourself manually.
 #'
+#' - `Any other user input`: the user will be asked again what's the desired action to be taken.
+#'
+#' The`APPROVE` and `DELETE` command will require a double confirmation given the criticality of the action.
```
caldwellst, line 38: I would add in `ARCHIVE`.

caldwellst (changes requested):
Overall looks good. Some linting/styling changes necessary that were caught in the linting that we have automatically going on in the repository. Then, just a few other minor code changes!

After that , you will need to increment the `.signals_version` file and update `CHANGES.md` before merging to main as well.

PR #253, opened 2024-11-04
Setup audience analysis

caldwellst:
Code used to do the audience analysis. I think you can just get rid of the plotting work (maybe just save in the Google Drive for posterity). Then just understand what the data frames generated are and do, and save those to the Azure cloud in a new folder, maybe something like `audience`. Shouldn't be very hard, and then all you need is to setup a weekly analysis on the Monday or something.


src/email/mailchimp/audience_analysis.R
```diff
@@ -0,0 +1,563 @@
...
+)
+
+box::use(
+  dplyr,
+  gg = ggplot2,
+  gghdx,
+  httr2,
+  lubridate,
+  purrr,
+  rnaturalearth,
```
zackarno, line 13: `{rnaturalearth}` is not a signals dependency (not in renv lockfile), may not be worth adding if it it won't be used in any pipeline

zackarno:
I'll just take this branch on. I may leave some comments on the code from seth mainly just as notes for myself

src/email/mailchimp/audience_analysis.R
```diff
@@ -0,0 +1,563 @@
...
+      activity_filters = "sent"
+    ) |>
+    httr2$req_perform() |>
+    httr2$resp_body_json() |>
+    purrr$pluck("activity") |>
+    length()
+}
+
+# get information on members
+df_members <- purrr$map(
```
zackarno, line 65: i think this `df_members` is the dataset we want to pipeline. The rest of analysis can come from it

zackarno:
@martinig94  -- I added a GHA here: https://github.com/OCHA-DAP/hdx-signals/actions/workflows/user_audience_analysis.yml.

- This currently creates a data.frame with user-analytics data and saves it to a `csv` on the `dev` blob. Once this is merged to main it should run every Monday at 8AM. 
- The csv is on the blob from GHA [run](https://github.com/OCHA-DAP/hdx-signals/actions/runs/11730347632) I did manually though `workflow_dispatch` (`HS_LOCAL= FALSE`).
- I guess we should check w/ Ed if this what there team needs next week?

### Remaining decisions

- what to do w/ `src/email/mailchimp/audience_analysis.R` file. 
    - The important part of that code to create the df/csv was scraped out and put in `src/email/mailchimp/create_user_analytics_dataset.R`, the remaining code is mostly to make the plots. 
    - Seth suggested deleting it and saving code to gdrive for posterity. This makes sense as it is more of an adhoc script with plots that does not fit into the overall system/architecture. This being said, I feel like there is a high probability someone will want us to re-run that analysis and update the plots at some point (soon) and that will be annoying to dig up and run. 
    - I guess ideally there would be a place in the repo for adhoc notebooks where we could keep this, but i appreciate that neither this or it's current script location really aligns with the repo structure as we have it now. My inkling would just be to create this `adhoc` folder in root and put this script there to avoid extra time on updating these plots in the future. We can always restructure and delete the folder at a later point

martinig94:
Hey @zackarno,
Apologies for the late reply, I missed your comment. Yes I agree in creating an `adhoc` folder for analysis that needs to be run _once in a while_ without removing the file from the repo.

src/email/mailchimp/create_user_analytics_dataset.R
martinig94, line 1: Things look clear and fine to me! Maybe let's take 10' during our call today to review the output file.

martinig94 (commented):
Things look clear and fine to me! Maybe let's take 10' during our call today to review the output file.

zackarno:
I was thinking about how to best log changes to database. I was going to just read in the new data set and compare to old. If anything new in old append it on. However, I noticed an example where the user changed `iso2`. Wonder how this should be reflected. Since we don't really know how this data will be used i think we should just write out a new file (csv) 1x per week and from that we can always retroactively merge when we understand exactly what's needed?

src/email/mailchimp/create_campaign_analytics_dataset.R
martinig94, line 1: Amazing!

src/email/mailchimp/create_user_analytics_dataset.R
martinig94, line 1: Cool! Thanks Zack!
Just reflecting a bit, in this case we would append new rows every time:
- a user modifies its interests
- there is an update in "open rate", "click rate" or email "received"
That's correct?
In this way, we might end up saving everything or almost every week. Do you agree?
I am thinking if there is a smarter way to proceed, what do you think?. Maybe topic for discussion tomorrow

.github/workflows/user_audience_analysis.yml
martinig94, line 1: Maybe let's just raname the file as `audience_analysis`? As we generate both campaign and user analytics in this GHA

src/email/mailchimp/campaign_analysis.R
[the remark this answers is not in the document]
zackarno: yup its been added to adhcoc.

I wonder if the structure of the email folder be unnested a bit to make it more clear which scripts are being sourced directly by GHA: i.e

```
|- email
    |- components
        |- etc. 1
        |- ect. 2
    |- mail chimp
        |- etc. 1
        |- etc. 2
    |- create_user_analytics.R
    |- create_campaing_analytics.R
````

File: .github/workflows/user_audience_analysis.yml
Comment (zackarno) [reply .github/workflows/user_audience_analysis.yml:1]: annoyingly the `.yml` file on this branch has to match that on `main` or else it cannot be run on this branch. Therefore I say we merge as is.. then do a small PR to tweak the name on `main`

File: src/email/mailchimp/create_user_analytics_dataset.R
Comment (zackarno) [reply src/email/mailchimp/create_user_analytics_dataset.R:1]: yeah agree. As discussed we now have an output specifically for: `hdx_signals_user_interactions.csv` that can reflect these updates without making other duplications in `hdx_signals_user_interests.csv`

File: src/email/mailchimp/campaign_analysis.R [hunk 3]
```diff
@@ -0,0 +1,131 @@
...
+
+# get all campaigns as a list
+campaigns <- base_api$mc_api(
+  lists_api = FALSE
+) |>
+  httr2$req_url_path_append(
+    "campaigns"
+  ) |>
+  httr2$req_url_query(
+    list_id = "e908cb9d48",
```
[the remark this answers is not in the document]
zackarno: i think it's not campaign specific but rather "project" specific... lol Mail Chimp probably made up a new term for "projects", but thats what I'm calling it. CHD has various other projects beyond Signals with there own ideas.

Nonetheless, good idea, i pulled it out and put the var at the top of the script

.github/workflows/monitor_acled_conflict.yaml
martinig94, line 1: to be moved in a separate branch (:

src/email/create_campaign_analytics_dataset.R
```diff
@@ -0,0 +1,138 @@
+
+FP_CAMPAIGN_DATA <- "output/user_research/campaign_analytics_data.csv"
```
martinig94, line 2: small thing: maybe for consistency with the other files let's call the file "hdx_signals_campaign_analytics_data"?

src/email/create_user_analytics_dataset.R
```diff
@@ -0,0 +1,303 @@
...
+  file_path = FP_USER_INTERESTS,
+  storage_account = STORAGE_ACCOUNT,
+  run_date = RUN_DATE
+  )
+
+
+
+#
+#
+# # User Interests ----------------------------------------------------------
```
martinig94, line 192: Can we get rid of the comemnted part?

src/email/create_user_analytics_dataset.R
```diff
@@ -0,0 +1,303 @@
...
+Script queries mailchimp API to get user analytics data and then combines it
+into a tidy data.frame.
+
+This will be run on a GitHub Action (.github/workflows/user_audience_analysis.yml)
+Weekly.
+"
+
+# putting this up here so that we can change to PROD later and only need
+# to update here rather than multiple places.
+FP_USER_INTERESTS = "output/user_research/hdx_signals_user_contacts.csv"
```
martinig94, line 13: `FP_USER_INTERESTS` and `FP_USER_CONTACTS` are swapped (:

martinig94 (changes requested):

src/email/create_user_analytics_dataset.R
zackarno: good call! - fixed

src/email/create_campaign_analytics_dataset.R
zackarno: sure! 
52f49f0

.github/workflows/monitor_acled_conflict.yaml
zackarno: indeed  - was just testing and i think it worked?

anyways reverted it back here: [88915fb](https://github.com/OCHA-DAP/hdx-signals/pull/253/commits/88915fb2ffd59778a5adce89004e3b7062ecd9b7)

src/email/create_user_analytics_dataset.R
```diff
@@ -0,0 +1,186 @@
...
+Weekly.
+"
+
+# putting this up here so that we can change to PROD later and only need
+# to update here rather than multiple places.
+FP_USER_INTERESTS = "output/user_research/hdx_signals_user_interests.csv"
+FP_USER_INTERACTIONS = "output/user_research/hdx_signals_user_interactions.csv"
+FP_USER_CONTACTS <- "output/user_research/hdx_signals_user_contacts.csv"
+
+STORAGE_ACCOUNT <- "dev"
```
martinig94, line 17: shall we move to prod?

martinig94 (approved):
maybe we need to save the files to 'prod'?

src/email/create_user_analytics_dataset.R
zackarno: sure why not! [02971f6](https://github.com/OCHA-DAP/hdx-signals/pull/253/commits/02971f668d1bce94decad27da780552d4a36f45c)

PR #267, opened 2024-11-21
Gha retry on error

zackarno:
WIP to see if we can add `action` to GHA that forces a retry on error.


zackarno:
@martinig94 - it looks like this code is working in this made up example. You can see the successful GHA output [here](https://github.com/OCHA-DAP/hdx-signals/actions/runs/11961816342).

The question would be:
1. on what scripts do we want to apply it to start out with.
2. what should we set the `timeout_minutes` parameter for

martinig94:
> @martinig94 - it looks like this code is working in this made up example. You can see the successful GHA output [here](https://github.com/OCHA-DAP/hdx-signals/actions/runs/11961816342).
> 
> The question would be:
> 
> 1. on what scripts do we want to apply it to start out with.
> 2. what should we set the `timeout_minutes` parameter for

Hey @zackarno,
That's great, thanks a lot!
I would apply it to the GHA that call the various APIs so all the `Monitor ...` GHA.
As `timeout_minutes` I would set it to 2 if we are going to have 2 reruns or to 5 if only one.

martinig94 (approved):
Amazing! Thanks a ton!

zackarno:
FYI - ran through the `monitor_*` GHA with the new yaml retry action. They all look fine. Maybe good for you to also have a look at that @martinig94

PR #270, opened 2024-11-25
add ind_module$ to indicator_id call in check_existing_signals

### LINKED ISSUE #271: GHA Failing: `indicator_id`not found
All Actions failed this morning with error:

```
Error in signals$generate_signals(ind_module = module, dry_run_filter = c("AFG",  : 
  object 'indicator_id' not found
Calls: <Anonymous> ... <Anonymous> -> filter.data.frame -> dplyr_quosures -> quos
```

Here is one of the [failed scheduled runs](https://github.com/OCHA-DAP/hdx-signals/actions/runs/12010265671) for further details

zackarno:
I think I found the source of the error. 

needed to attach the `ind_module` namespace to to the indicator_id inside of:

```
  check_existing_signals$check_existing_signals(
    indicator_id = indicator_id,
    fn_signals = fn_signals
  )
```

this check gets bypassed when `HS_DRY_RUN=TRUE` which is why we didn't see it before.

Lessons learned - let's add more logging!

closes #271


zackarno:
`ind_module` module namespace also had to be added to all instances where `indicator_id`. I fixed throughout `generate_signals()` func and now the [GHA run](https://github.com/OCHA-DAP/hdx-signals/actions/runs/12017112437) was successful w/ `HS_LOCAL` and `HS_DRY_RUN` both equal to `FALSE` as it will be for scheduled runs. Therefore, we should be good to go?

martinig94 (approved):

PR #274, opened 2024-11-27
remove interaction data from interest data before appending

zackarno:
added step to remove user interaction data from user interests data.frame prior to checking blob and appending data. I deleted user-interests on blob to start fresh csv with this [run](https://github.com/OCHA-DAP/hdx-signals/actions/runs/12054973407)....

Do we forsee an issue that now we have one extraction date for user interests...but multiple for others? 

i think we could ignore the CHANGES failed test for this

### DIFF
--- a/src/email/create_user_analytics_dataset.R
+++ b/src/email/create_user_analytics_dataset.R
@@ -137,6 +137,13 @@ df_members <- purrr$map(
     name != "" # HDX Signals email
   )
 
+df_member_interactions_new <- df_members |>
+  dplyr$group_by(id) |>
+  dplyr$distinct(open_rate, click_rate, email_count) |>
+  dplyr$ungroup() |>
+  dplyr$mutate(
+    extraction_date = lubridate$as_date(Sys.Date())
+  )
 
 df_members_new <- df_members |>
   dplyr$mutate(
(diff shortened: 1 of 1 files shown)


martinig94:
> added step to remove user interaction data from user interests data.frame prior to checking blob and appending data. I deleted user-interests on blob to start fresh csv with this [run](https://github.com/OCHA-DAP/hdx-signals/actions/runs/12054973407)....
> 
> Do we forsee an issue that now we have one extraction date for user interests...but multiple for others?
> 
> i think we could ignore the CHANGES failed test for this

Hey Zack! @zackarno 

Agree on the changes test, does not really matter in this case!

What do you mean by "ne extraction date for user interests...but multiple for others"?
We should have one extraction date per user-interest per user if it does not change in time.

Example:
Let's assume that I subscribed the first of September to everything: We will have a dataset with n rows as many are the interests, all of them will have the 'interested' column = to TRUE and the extraction date would be the first date in which we started retrieving this info. 
If today I modify my subscription and I will receive infomation for everything except for the conflicts in AFG, next monday in the dataset there will be only one additional raw for my user and related to AFG - conflict. This time the column 'interested' will be equal to FALSE and the extraction date will be next monday. Let me know if it's clear and if you agree with the logic proposed. Otherwise happy to jump in a call (tomorrow) if some additional brainstorming is needed (:

zackarno:
hey @martinig94  i think you are right and i was over thinking it! Therefore, you good for us to merge it. I just pushed a something to fix the lint fail.

martinig94 (approved):
Looks good to me! I did not run the code to see if the output is as expected. Did you or do you want me to do it?

zackarno:
great will merge - yeah i ran it manually [here](https://github.com/OCHA-DAP/hdx-signals/actions/runs/12054973407)

PR #294, opened 2025-05-06
global cholera data exploration

PaulineNimo:
@martinig94 do you mind reviewing this? It is just the R script I have added here. Thanks!


martinig94:
Hi @PaulineNimo,
I think it looks fine!
if you can fix the two failing tests and address my only minor comment I think we can merge it (:
Let me know if you need any help, 
Thanks!

Issue #303, opened 2025-08-05
Azure permissions error


hannahker:
On triaging the latest signal, I got the following error: 

```
> triage_signals$triage_signals("idmc_displacement_conflict",0, test=FALSE)
Tell us what you want to do with the following commands:

APPROVE: Send campaigns and add to `output/signals.parquet`
DELETE: Delete the campaign content and `output/idmc_displacement_conflict/signals.parquet` file
so you can recreate later.
ARCHIVE: Delete the email campaign, but move the alert to `output/signals.parquet`
DO_NOTHING: Do nothing, so you can decide later.
Your command: APPROVE
You typed the command: APPROVE
Given the criticality of the action please type I CONFIRM to proceed with the action selectedYour command: I CONFIRM
Error in process_storage_response(response, match.arg(http_status_handler),  : 
  Forbidden (HTTP 403). Failed to complete Storage Services operation. Message:
AuthorizationPermissionMismatch
This request is not authorized to perform this operation using this permission.
RequestId:95a79ee2-601e-0002-4715-06ab38000000
Time:2025-08-05T14:33:03.1145673Z.
```

However the signal sent properly.

PR #305, opened 2025-08-06
hotfix bug: change  to  for ARCHIVE modality

zackarno:
tiny typo - not sure if it is involved in our latest sequence of errors, but should be corrected nonetheless

https://github.com/OCHA-DAP/hdx-signals/issues/303
https://github.com/OCHA-DAP/hdx-signals/actions/runs/16781044820


hannahker (approved):
Good catch!

PR #344, opened 2026-07-08
Feat: add global choropleth map for sea5_anomaly indicator

### LINKED ISSUE #343: Add map to sea5_anomaly indicator
For the sea5_anomaly indicator there won't be a plot but only a global map which should look like the country level map in
https://ocha-dap.github.io/ds-seas5-skill/#map. The code producing this map is in https://github.com/OCHA-DAP/ds-seas5-skill
An example of the map is also attached
Image

the map coloring depends on the input dataframe of the map function. Use similar name conventions as the other indicators (probably df_raw should be the name).
The Map should display the next valid trimester. Example: if it's the july forecast the map should display the ASO (August-September-October) values

martinig94:
## Summary

Closes #343

- Replaces the per-country adm0 boundary map with a global choropleth map showing `p_5rp` values for all countries
- The map displays the next valid trimester (e.g., July forecast → ASO), computed from the alert date
- Countries with no data are shown in grey; coloring uses a Yellow-Orange-Red scale from 0–100% probability
- Switched from `settings = "map"` (per-country dimensions) to `settings = "plot"` with fixed 10x5 inch global dimensions

## Test plan

- [ ] Trigger a dry run of the `monitor_sea5_anomaly` workflow and verify the map image shows a world map with countries colored by `p_5rp`
- [ ] Confirm trimester label in the map title is correct for the forecast month (e.g., July forecast → `ASO`)
- [ ] Confirm countries with no data display in grey

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/indicators/sea5_anomaly/utils/map_sea5_anomaly.R
+++ b/src/indicators/sea5_anomaly/utils/map_sea5_anomaly.R
@@ -1,8 +1,7 @@
 box::use(
   dplyr,
   gg = ggplot2,
-  rnaturalearth,
-  scales
+  rnaturalearth
 )
 
 box::use(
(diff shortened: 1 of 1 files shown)


martinig94:
Forecast Map
Consider updating legend to use one contiguous color strip from drier to wetter and another strip below for skill showing the solid/hatched patterns -- use brand color tokens from redesign components page
Strongly below: #7f5619
Below: #dda555
Roughly: #e2e8e8 
Above: #74a1e8
Strongly above: #134ead
Outside rainy season: #b1c1c2
Not monitored: #f5f7f7

martinig94:
Done in 3b50de6: the anomaly legend is now one contiguous strip from drier to wetter using the brand hex values you listed (strongly below `#7f5619` → strongly above `#134ead`, plus outside rainy season `#b1c1c2` and not monitored `#f5f7f7`). Forecast skill is now a separate legend below it, shown as a solid/stripe/crosshatch pattern overlay (via `ggpattern`, already available in the lockfile) instead of being folded into extra color categories: solid = high skill, light stripe = moderate skill, crosshatch = low skill.

Pushed to this branch — ready for another look.

PR #348, opened 2026-07-22
Restyle Signals charts, maps, and email to 2025 HDX redesign

martinig94:
Closes #347

## Summary

Applies the HDX website 2025 redesign guidance from #347 (Merriweather display titles, Roboto body copy, primary blue `#1862D8`, institutional data-grid blue `#4681E0`, cool-grey hairlines) to:

- **Line/fatalities charts** (`plot_ts()`, used by ACLED conflict, IDMC displacement, cholera): primary-blue line, translucent area fill under the line, and a haloed endpoint dot, per the "Fatalities Line Chart" design file.
- **Agricultural hotspot matrix/heatmap** (`plot_agricultural_hotspots.R`): institutional data-grid blue cells, hairline grid, and a new focus ring around the declared hotspot month, per the "Hotspot Chart Restyle" design file.
- **Shared theme** (`theme_signals.R`): loads Merriweather + Roboto and applies them to titles/body text, hairline gridlines and axis ticks, muted axis text color.
- **Mailchimp email template** (`hdx_signals_template.html`, `banner_block.R`): header/banner background and link color switched to the new primary blue, `<h1>` set to Merriweather, body text to Roboto — matching the annotated PDF in the issue.
- **Point maps** (`map_points.R`, `map_conflict.R`, `geom_cities.R`, `map_theme.R`): restyled per the correct "HDX data-viz style guide" link added in the issue comments (the original "Signals map" link was a duplicate of the line-graph link). The event symbol is now a single translucent-fill + solid-stroke marker (was two stacked layers), country fill/boundary use the new neutral tokens, place labels/legend text switched to Roboto/Roboto Mono, and the ACLED conflict map (fatalities-specific) now uses the style guide's danger red instead of the neutral blue.
- **Market monitor bar chart** (`plot_market_monitor.R`): same primary-blue/hairline token swap for visual consistency, since it used the exact old tokens (`sapphire-hdx`, `gray-dark`) being replaced elsewhere — no dedicated design was provided for it.

New file `src/images/plots/hdx_signals_palette.R` holds the new hex tokens, since the pinned `gghdx` package (v0.1.4, GitHub `main`) only implements the old sapphire/mint/tomato/gray palette and Source Sans 3 — it hasn't been updated for this redesign yet. Chart-side changes are therefore local overrides layered on top of `gghdx`'s theme, not a change to `gghdx` itself.

## Flagged — could not fully address / judgment calls made

[...]

PR #350, opened 2026-07-29
Export latest three signals to Google Drive after triage

### LINKED ISSUE #349: Export the three most recent signals to Google Drive after triage
## Problem

Right now the only durable output of triaging a signal is `output/signals.parquet` (internal) and `output/signals.csv` (pushed to HDX via the `hdx-signals-alerts` pipeline). There's no lightweight, easy-to-poll feed of just the latest signals for external tools that don't want to pull the full HDX dataset or hit the Azure blob store directly.

## Request

Every time a signal is triaged with `APPROVE` or `ARCHIVE` in `triage_signals()` (`src/signals/triage_signals.R`), generate a small file from `output/signals.csv` containing only the three most recently dated signals, and store it on Google Drive so it can be shared/consumed outside the usual Azure/HDX pipeline.

## Proposed approach

- Hook into `save_core_signals_hdx()`, which already builds and writes the exact data frame saved to `output/signals.csv` on every `APPROVE`/`ARCHIVE` triage action (both routes call it) — reuse that data frame directly rather than re-reading the CSV back from Azure.
- Take the 3 rows with the most recent `date` value.
- Upload as a CSV to a Google Drive folder via a service account (no interactive OAuth, since this runs in GitHub Actions), following the existing `cloud_storage$update_az_file()` pattern (skip upload when `HS_LOCAL` is `TRUE`).
- New env vars: `HS_GDRIVE_SERVICE_ACCOUNT` (service account JSON) and `HS_GDRIVE_FOLDER_ID` (destination folder ID), documented in `ENVIRONMENT.md`.
- Overwrite the same file in place on each run (rather than creating a new file every triage) so downstream consumers can poll one stable link.

## Open questions for reviewers

- Is `date` (the alert date) the right recency column, or should this use `campaign_date`/`triage_time` (when it was actually triaged) instead?
- Is a single shared service account + folder the right access model, or should this go through a different sharing mechanism?

martinig94:
Fixes #349

## Summary

Stacked on #348 (branched from its head so this can be reviewed independently; base should be retargeted to `main` once #348 merges).

- **`src/utils/push_google_drive.R`** (new): authenticates to Google Drive non-interactively via a service account (`HS_GDRIVE_SERVICE_ACCOUNT`, raw JSON, never written to disk), writes a data frame to a temp CSV, and uploads it to a Drive folder (`HS_GDRIVE_FOLDER_ID`). If a file with the same name already exists in that folder it's updated in place (`drive_update`) rather than duplicated, so the same Drive link always has the latest content. Skips entirely when `HS_LOCAL` is `TRUE`, matching `cloud_storage$update_az_file()`.
- **`src/signals/triage_signals.R`**: `save_core_signals_hdx()` now calls a new `save_latest_signals_gdrive()` after writing `output/signals.csv` and pushing to HDX. It reuses the exact data frame just saved to `signals.csv`, takes the 3 rows with the most recent `date` (`dplyr::slice_max()`), and uploads them as `hdx_signals_latest.csv`. This runs on every `APPROVE` or `ARCHIVE` triage command, since both route through `save_core_signals_hdx()`.
- Added `googledrive`/`gargle` to `renv.lock` — all of their transitive deps (`httr`, `fs`, `rlang`, etc.) were already present at compatible versions, so this is a pure two-package addition, nothing else touched or bumped.
- `ENVIRONMENT.md`: documented the two new env vars.
- `.github/workflows/triage_signals.yaml`: wired `HS_GDRIVE_SERVICE_ACCOUNT`/`HS_GDRIVE_FOLDER_ID` through from repo secrets (still need to be added as actual GitHub secrets before this can run for real).
- `src/utils/__tests__/test-push_google_drive.R`: unit tests for the `hs_local()` skip path, update-existing-file path, and upload-new-file path, mocking all Drive/network calls (mirrors `test-cloud_storage.R`'s style).
- Version bumped `0.6.0.0` → `0.7.0.0`, `CHANGES.md` updated.

## Judgment calls (flagging per issue #349's open questions)

1. **Recency column**: used `date` (the alert date), not `campaign_date`/`triage_time` (when triaged). `date` is what `output/signals.csv` is conceptually ordered by for consumers, but happy to switch if `campaign_date` is preferred.
2. **Overwrite vs. new file per run**: chose overwrite-in-place (`hdx_signals_latest.csv`, one stable Drive file/link) over creating timestamped files each triage, since "latest three signals" reads as a live feed rather than a history log. No versioning/history is kept on Drive as a result — only Azure/`signals.parquet` retains full history.
3. **Auth model**: single shared service account + single destination folder, consistent with how `DSCI_AZ_BLOB_PROD_SAS_WRITE` is a single shared credential. Not tested against a real Drive folder — no credentials available in this environment (same limitation flagged in #348 for Mailchimp).

## Test plan

- [x] `parse()`-validated both edited/new R files
- [x] `Rscript src/utils/__init__.R` (full `src/utils` suite): 170 passing, only 3 pre-existing failures in `test-track_summary_input.R` caused by a Windows file-locking quirk unrelated to this change (confirmed by reproducing the same 3 failures on the unmodified branch)
- [x] `lintr::lint()` on both new/changed files: no lints
- [ ] Real Google Drive upload — no service account credentials available in this environment to test against an actual Drive folder


martinig94:
Folder ID pinned to the one you shared (hardcoded in `src/utils/push_google_drive.R`, not a secret — matches the `MC_SIGNALS_LIST_ID` pattern already used for Mailchimp). Only `HS_GDRIVE_SERVICE_ACCOUNT` (the service-account JSON) is still needed as an actual repo secret for this to run for real.

martinig94:
## Status: paused, blocked on Google Drive access

Code and unit tests are in good shape (all passing, no lints), but real end-to-end verification against an actual Drive folder is blocked:

- A regular "My Drive" folder doesn't work with a bare service account — Google returns a hard 403 on upload (`Service Accounts do not have storage quota... Leverage shared drives`), since service accounts have no personal storage quota outside a Shared Drive.
- The original Shared Drive folder 404s for the service account, and the org's policy prevents adding it as a Shared Drive member (external/cross-project accounts appear to be blocked from being added).

Unblocking this needs one of:
1. A Workspace admin either allows the service account onto the Shared Drive, or sets up domain-wide delegation so it can impersonate a real account.
2. Switch the auth model entirely to OAuth as a real user account instead of a service account (bigger change: one-time interactive login to cache a token, then store that cached token as a secret for the Action to reuse non-interactively).

Also flagging: during local debugging, the service account's private key was inadvertently printed in a local terminal (via an `.Renviron` parse-error dump) and pasted into this conversation. **The key should be rotated** before this goes further, regardless of which auth path we take.

Leaving this PR open as-is until there's a path to test real Drive access.

